PR #5, opened 2020-10-12
Add make_nanopub_keys command to setup RSA keys for nanopub

svenvanderburg:


raar1 (commented):
OK, so this looks good to me. I also see that PR #7 is a better place for any comments I have on the code refactoring.

My only question about this PR concerns the issue of using RSA keys to sign and retract nanopublications. So it appears that the expected way to 'author' a nanopub is by publishing an introductory nanopub (with e.g. your ORCID or something like that). The URI to that nanopub can then be used as the author when using the ```publish()``` method of the client. However, the nanopubs are also signed using whatever RSA key ```np``` is configured to use (the Trusty URI stuff fits in here). In some sense, this is 'real' proof of who actually made the nanopub. It also allows the person with the key (and only that person) to retract nanopubs that they have published. That is not in the client yet, but we will presumably add it at some point. I've made an issue for it: #8 .

All of this leads me to think that making people new keys should be carefully controlled, since a user very likely wants to use the same key (and introductory nanopub) that they normally use. Would it then make sense to have some sort of ```NanoUserProfile``` class? Or a ```nano_user_profile.yml``` or something like that, which would contain the introductory nanopub and the path to the RSA keys they want to use?

I would then like, perhaps, to force a user to explicitly load the user profile they want to act as, perhaps at the start of the script. I guess we don't need this for searching and fetching, but if the user wants to ```publish()``` then perhaps they should need to provide an ```author=``` kwarg that gives this user info? I've made an issue for that too: #9 .

PR #7, opened 2020-10-13
Refactor

svenvanderburg:
Refactor Nanopub class to NanopubClient, make the class methods into proper methods.
Refactor NanopubObj class to Nanopub.
Move the functionality of the `to_rdf` method to the `from_assertion` classmethod of the new `Nanopub` class.
Move namespaces to a separate namespaces module.

It think there is still a lot to improve, but these are I think the major refactorings that we want to do.

### DIFF
--- a/nanopub/nanopub.py
+++ b/nanopub/nanopub.py
@@ -271,7 +271,8 @@ class NanopubClient:
         return np_rdf
 
     @staticmethod
-    def publish(assertionrdf, uri=None, introduces_concept=None, derived_from=None, attributed_to=None, nanopub_author=None):
+    def publish(self, assertionrdf, uri=None, introduces_concept=None, derived_from=None,
+                attributed_to=None, nanopub_author=None):
         """
         Publish the given assertion as a nanopublication with the given URI.
         Uses np commandline tool to sign and publish.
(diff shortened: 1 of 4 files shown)


nanopub/nanopub.py
```diff
@@ -8,84 +8,198 @@
 import requests
 from rdflib.namespace import RDF, DC, DCTERMS, XSD
 
-from nanopub import java_wrapper
+from nanopub import java_wrapper, namespaces
+
+DEFAULT_URI = 'http://purl.org/nanopub/temp/mynanopub'
 
 
 class Nanopub:
```
svenvanderburg, line 16: Now that I added the `from_assertion` method, this `Nanopub` class is starting to feel a lot more like a wrapper for RDF... That rings a bell right? We could subclass the `RdfWrapper` class from fairworkflows, or even move the Nanopub class there (although that seems kind of strange). Preferably we should do that in a separate issue. What do you think?

svenvanderburg:
When this is released I will update fairworkflows to also match this new structure.

nanopub/nanopub.py
raar1: This is a good point. Although given that the nanopub lib will be used in a number of non-fairworkflows related places as well (hopefully) then maybe it's still better that it just return the raw ```rdflib``` graph, rather than wrapping it all up. However, I am also wondering now... are there any other libraries out there that provide a powerful class that looks like ```RdfWrapper``` that we could use for both libs?

svenvanderburg: OK, we'll just stick with this for now. Good point about existing libraries having some sort of `RdfWrapper`. I did a quick search (for example [https://wiki.python.org/moin/RdfLibraries](here)), but couldn't really find anything.

raar1 (approved):

PR #15, opened 2020-10-23
Add better error handling when calling java tool

svenvanderburg:
* Add better error handling when calling the nanopub java tool
* Refactor java_wrapper functions into a JavaWrapper class


raar1 (approved):
Introduction of ```JavaWrapper``` class and use of subprocess is a big improvement.

PR #16, opened 2020-10-23
Make introduces_concept property of Nanopub object reflect the nanopub pubinfo graph

svenvanderburg:


raar1 (approved):
This is a good change. I like that it handles more than one concept being introduced by a single nanopub.

PR #21, opened 2020-10-26
Multiple server endpoints

svenvanderburg:
- Added retry to different grlc servers if one of them is down.
- Also refactored the searching code quite a lot.


nanopub/nanopub.py
```diff
@@ -200,114 +201,117 @@ class NanopubClient:
...
-                nanopubs.append(nanopub)
+            nanopub['date'] = result['date']['value']
 
-                if len(nanopubs) >= max_num_results:
-                    break
+            nanopubs.append(nanopub)
 
-            return nanopubs
+            if len(nanopubs) >= max_num_results:
+                break
```
svenvanderburg, line 311: I didn't change anything here, github is just confused.

tests/test_nanopub.py
```diff
@@ -9,96 +9,85 @@
 from nanopub import NanopubClient, namespaces
 from nanopub.nanopub import Nanopub
 
-DEFAULT_FORMAT = '.trig'
-BAD_GATEWAY = 502
-NANOPUB_SERVER = 'http://purl.org/np/'
-SERVER_UNAVAILABLE = 'Nanopub server is unavailable'
+skip_if_nanopub_server_unavailable = (
+    pytest.mark.skipif(requests.get('http://grlc.nanopubs.lod.labs.vu.nl/').status_code != 200,
```
svenvanderburg, line 13: This should be replaced with the test server later.

nanopub/nanopub.py
raar1: Ha, I'll trust you... this time... :)

nanopub/nanopub.py
```diff
@@ -11,6 +12,12 @@
 from nanopub import namespaces
 from nanopub.java_wrapper import JavaWrapper
 
+NANOPUB_GRLC_URLS = ["http://grlc.nanopubs.lod.labs.vu.nl/api/local/local/",
```
raar1, line 15: Just wondering if we should have this list in a config file, or something? I believe Tobias mentioned the possibility of having new nanopub servers announcing their existence via nanopubs, so we might want a method to scan for such nanopub servers, and add to the list (probably a task for another time). But I guess this list should be in a ```known_nanopub_servers``` file, or something?

nanopub/nanopub.py
```diff
@@ -200,114 +207,111 @@ class NanopubClient:
     """
     def __init__(self):
         self.java_wrapper = JavaWrapper()
+        self.grlc_urls = NANOPUB_GRLC_URLS
 
-    def search_text(self, searchtext, max_num_results=1000,
-                    apiurl='http://grlc.nanopubs.lod.labs.vu.nl//api/local/local/find_nanopubs_with_text'):
+    def find_nanopubs_with_text(self, text, max_num_results=1000):
```
raar1, line 212: I think this renaming has been a good idea. It makes everything consistent with the grlc API.

nanopub/nanopub.py
```diff
@@ -200,114 +207,111 @@ class NanopubClient:
...
-            # Make sure that results are provided
-            try:
-                results_json = r.json()
-            except:
-                # If the returned message can't be serialized as JSON (such as due to virtuoso error) then there are no results
-                print('Error: Could not serialize response as JSON:\n', r.content)
-                return []
+    def _search(self, endpoint, params, max_num_results):
+        """
+        General nanopub server search method. User should use e.g. search_text() or search_pattern() instead.
```
raar1, line 284: Update search_text etc to new names

nanopub/nanopub.py
```diff
@@ -200,114 +207,111 @@ class NanopubClient:
...
 
-            results_list = results_json['results']['bindings']
-            nanopubs = []
+        Raises:
```
raar1, line 286: Maybe include the arguments in the docstring, and what form they should take (I guess only ```params``` is likely to be unclear)

nanopub/nanopub.py
```diff
@@ -200,114 +207,111 @@ class NanopubClient:
...
-
-        if searchparams is None:
-            raise ValueError('kwarg "searchparams" must be specified. Consider using search_text() function instead.')
-
-
-        # Query the nanopub server for the specified text
         headers = {"Accept": "application/json"}
-        r = requests.get(apiurl, params=searchparams, headers=headers)
+        r = None
+        for grlc_url in self.grlc_urls:
```
raar1, line 270: Do we always try the first server in the list, or should it start from one at random? Perhaps always following the order in the list is better behaviour, even though it will mean one server always being loaded more than the others...?

tests/test_nanopub.py
raar1: For testing publishing, yes, although for simple fetching of existing nanopubs at known URIs it might be worth still using the real servers for that, just to indicate to us if something there has changed.

tests/test_nanopub.py
```diff
@@ -34,91 +34,78 @@ def test_nanopub_construction_with_bnode_introduced_concept():
...
-        client._search(searchparams={'search': 'text'},
+        client._search(params={'search': 'text'},
                        max_num_results=100,
-                       apiurl=None)
+                       endpoint=None)
 
 
 @pytest.mark.flaky(max_runs=10)
-@pytest.mark.skipif(nanopub_server_unavailable(), reason=SERVER_UNAVAILABLE)
+@skip_if_nanopub_server_unavailable
```
raar1, line 104: This is much cleaner/nicer than before :)

raar1 (approved):
Good PR. Very useful changes, and I'm glad we will no longer be dependent on a single server. I've approved, but left one or two comments. Putting the server URIs in a config file could be done in this PR, or feel free to make an issue and we can do it later (not urgent).

nanopub/nanopub.py
svenvanderburg: I am in favour of just leaving it like this (which is the simplest option in my mind). If at some point we indeed have an automatic process of keeping the list of nanopub servers we can change it into a config file.

Also, I think this is a problem that should ideally be fixed on the nanopub server side. They should just have a proxy that routes traffic to one of the servers that is least loaded (load balancing). Users (like us) should not have to worry about this.

nanopub/nanopub.py
svenvanderburg: Great!

nanopub/nanopub.py
svenvanderburg: Check

nanopub/nanopub.py
svenvanderburg: Yes, let's assume we get huge traffic through our client (aim for the moon 😛 ). I will change it to pick a random one.

tests/test_nanopub.py
svenvanderburg: Let's discuss this on the test endpoint PR.

nanopub/nanopub.py
svenvanderburg: Check!

tests/test_nanopub.py
svenvanderburg: Great :)

PR #25, opened 2020-10-28
Use test endpoint

svenvanderburg:


tests/test_java_wrapper.py
```diff
@@ -49,3 +50,11 @@ def test_sign_nanopub_no_rsa_key(tmp_path):
     with pytest.raises(RuntimeError) as e:
         java_wrapper.sign(unsigned_file=temp_unsigned_file)
     assert 'RSA key appears to be missing' in str(e.value)
+
+
+@pytest.mark.flaky(max_runs=10)
+@skip_if_nanopub_server_unavailable
```
svenvanderburg, line 56: This will actually check if garlic server is available, I would propose to just leave this for now to reduce complexity.

tests/test_nanopub.py
```diff
@@ -1,19 +1,13 @@
...
+from nanopub import NanopubClient, namespaces, Nanopub
 
-skip_if_nanopub_server_unavailable = (
-    pytest.mark.skipif(requests.get('http://grlc.nanopubs.lod.labs.vu.nl/').status_code != 200,
-                       reason='Nanopub server is unavailable'))
-
-client = NanopubClient()
+client = NanopubClient(use_test_server=True)
 
 def test_nanopub_construction_with_bnode_introduced_concept():
```
svenvanderburg, line 12: @raar1 you mentioned on #21: For testing publishing, yes, although for simple fetching of existing nanopubs at known URIs it might be worth still using the real servers for that, just to indicate to us if something there has changed.

svenvanderburg: I would propose to always use test server when testing. The nanopub test server should actually be the first server to undergo a change, before the real servers change. (but if you think in practice this will not happen we should indeed change back to testing the real servers).

nanopub/java_wrapper.py
```diff
@@ -33,7 +43,20 @@ def sign(self, unsigned_file: Union[str, Path]) -> str:
...
+        Publish the signed nanopub to the nanopub server. Publishing to the real server depends
+        on nanopub-java, for the test server we do a simple POST request.
+
+        TODO: Use nanopub-java for publishing to test once it supports it.
+        """
+        if self.use_test_server:
+            headers = {'content-type': 'application/x-www-form-urlencoded'}
+            with open(signed, 'rb') as data:
+                r = requests.post(NANOPUB_TEST_SERVER, headers=headers, data=data)
+            r.raise_for_status()
```
svenvanderburg, line 57: I don't really like this, but we don't have any other options for now.

tests/test_nanopub.py
svenvanderburg: Also, if we add support for more search endpoints we can populate the test server with stuff we expect to be returned by searches without polluting the test server.

nanopub/java_wrapper.py
```diff
@@ -4,18 +4,28 @@
 from typing import Union
 
 import rdflib
+import requests
 
 from nanopub.definitions import PKG_FILEPATH
 
 # Location of nanopub tool (currently shipped along with the lib)
 NANOPUB_SCRIPT = str(PKG_FILEPATH / 'np')
+NANOPUB_TEST_SERVER = 'http://test-server.nanopubs.lod.labs.vu.nl/'
```
raar1, line 13: Wondering if we should have a central config file that specifies ```NANOPUB_TEST_SERVER```, ```NANOPUB_JAVA_VERSION``` etc, just so we can see at a glance how the library is configured? Or do you think it's better keeping the settings local?

nanopub/java_wrapper.py
raar1: Potentially we could consider taking publishing out of the java wrapper completely, and having the ```nanopub``` python lib do it itself using ```requests```. We would then keep ```nanopub-java``` for things like making nanopubs 'trusty' etc. Could that help to santize this a little?

raar1: i.e. publishing wouldn't require ```if self.use_test_server``` anymore, since you would always specify which URL to publish to yourself, rather than using the java tool

tests/test_java_wrapper.py
raar1: Agreed.

tests/test_nanopub.py
raar1: Yes, good point about the test server also being the first place that changes to the API should happen. Could it still be worthwhile keeping a single fetch test that also tries the existing servers? I'm thinking it could maybe help debugging if someone changes existing nanopub servers without it going via our test server first? But otherwise I'm happy enough having everything go via the test server for the reason you mentioned.

raar1 (approved):
Very useful to be able to use the test server - should cut down on the pollution ;)

I've left a couple of minor comments.

nanopub/java_wrapper.py
svenvanderburg: Yeah I think that makes sense. I used to have that in most projects I worked on. I will do that in a separate PR.

nanopub/java_wrapper.py
svenvanderburg: Yeah, could be a good option to always do the post ourselves. But doesn't nanopub-java do some magic to make sure something is published to all servers?

Preferably as a user you don't want to specify where to publish to. You shouldn't want to care about that.

I would propose to leave it like this (and leave the TODO in there to use nanopub-java again). Or would you prefer doing the POST from python? @raar1

tests/test_nanopub.py
svenvanderburg: Yeah, I guess that makes sense. I added one test for querying production server.

nanopub/java_wrapper.py
raar1: I'm happy to leave it as a TODO as you suggest. I also would like to avoid putting that functionality into python unless we know exactly how the java tool is doing it. It's not a big problem, so we can worry about this another time.

Issue #18, opened 2020-10-26
Fix nanopub-java version that we use 


svenvanderburg:
(we now automatically fetch latest)

raar1:
Using a docker image of the nanopub-java tool is something we should probably be doing, and is likely relevant to this issue.

svenvanderburg:
Yes that makes sense!

svenvanderburg:
There is only one docker image on [nanopub-java docker hub](https://hub.docker.com/r/umids/nanopub-java/tags) that is younger than the latest release of nanopub-java... So I implemented this instead: https://github.com/fair-workflows/nanopub/pull/27 (it just fixes the java .jar file that is being fetched instead of latest).

PR #27, opened 2020-10-30
Adapt np script to download nanopub-java version 1.32 instead of latest

svenvanderburg:
NB: Docker-based solution is undesired, hence this PR. (see discussion on https://github.com/fair-workflows/nanopub/issues/18)

### DIFF
--- a/nanopub/np
+++ b/nanopub/np
@@ -2,17 +2,12 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


raar1 (approved):
Yes, this seems like it could be a better way of fixing the dependency for now. We can revisit the docker thing at a later date if the situation changes.

PR #29, opened 2020-11-03
Issue 9 create user profile

dsmits:
Created a command-line interface for generating rsa keys and uploading a this-is-me nanopub.
I wanted to include more tests but it turned out to be a hell in combination with the click library.


dsmits:
I had issues with running unit tests on this script that is in the bin dir so I moved it to the root dir. I could have somehow added the bin dir to the pythonpath but I wasn't sure which approach would've been best. Any suggestions?

tests/test_setup_profile.py
```diff
@@ -0,0 +1,51 @@
...
+PUBLIC_KEYFILE = 'id_rsa.pub'
+PRIVATE_KEYFILE = 'id_rsa'
+NANOPUB_DIR = '.nanopub'
+
+ORCID = '123'
+NAME = 'pietje'
+PUBLISH = False
+
+
+def test_provided_keypair_copied_to_nanopub_dir(tmp_path: Path):
```
svenvanderburg, line 17: It feels a bit like the testing code is more complicated (and thus more difficult to maintain) than the code that you're testing. And it only covers a small part of setup_profile.py anyways. What do you think?

setup_profile.py
```diff
@@ -0,0 +1,100 @@
...
+              prompt=f'If the public and private key you would like to use are not in {USER_CONFIG_DIR}, '
+                     f'provide them here. If they are in this directory or you wish to generate new keys, '
+                     f'leave empty.',
+              help='Your RSA public and private keys with which your nanopubs will be signed',
+              default=None)
+@click.option('--orcid', type=str, prompt=True, help='Your ORCID')
+@click.option('--name', type=str, prompt=True, help='Your name')
+@click.option('--publish/--no-publish', type=bool, is_flag=True, default=False,
+              help='If true, nanopub will be published to nanopub servers',
+              prompt='Would you like to publish your profile to the nanopub servers?')
```
svenvanderburg, line 34:
```suggestion
              prompt=('Would you like to publish your profile to the nanopub servers?'
                               'this links your ORCID to your RSA key, thereby making all your'
                               'publications linkable to you'))
```
Or something like that.

setup_profile.py
```diff
@@ -0,0 +1,100 @@
...
+    # Construct your desired assertion (a graph of RDF triples)
+    my_assertion = Graph()
+
+    key_declaration = BNode('keyDeclaration')
+    orcid_node = ORCID[orcid]
+
+    my_assertion.add((key_declaration, NPX.declaredBy, orcid_node))
+    my_assertion.add((key_declaration, NPX.hasAlgorithm, Literal(RSA)))
+    my_assertion.add((key_declaration, NPX.hasPublicKey, Literal(public_key)))
+    my_assertion.add((orcid_node, FOAF.name, Literal(name)))
```
svenvanderburg, line 85: This looks good!

setup_profile.py
```diff
@@ -0,0 +1,100 @@
...
+from nanopub.java_wrapper import JavaWrapper
+
+PRIVATE_KEY_FILE = 'id_rsa'
+PUBLIC_KEY_FILE = 'id_rsa.pub'
+DEFAULT_PRIVATE_KEY_PATH = USER_CONFIG_DIR / PRIVATE_KEY_FILE
+DEFAULT_PUBLIC_KEY_PATH = USER_CONFIG_DIR / PUBLIC_KEY_FILE
+RSA = 'RSA'
+
+
+@click.command()
```
svenvanderburg, line 23: Maybe have a general help as well? I.e. waht does this command do?

svenvanderburg (commented):
Nice! Great work :) This is a really useful feature.
* Could you also implement storing the ORCID_ID and by default adding it to the provenance of a nanopub when publishing? Or do you plan to do that separately?
* We could think of some sort of validation for the ORCID ? But maybe we should do it in a separate issue.

svenvanderburg:
> I had issues with running unit tests on this script that is in the bin dir so I moved it to the root dir. I could have somehow added the bin dir to the pythonpath but I wasn't sure which approach would've been best. Any suggestions?

Hmmm, not sure. Was this locally or in a github action? Let's have a look together.

tests/test_setup_profile.py
dsmits: Yeah I know. I would vote for keeping it, because I think it is very important to check if the cli actually does the stuff it promises to do, so my view on it is that it is not enough testing.  If it becomes a pain in the ass to maintain it we can think of something.

In terms of maintainability I think it wouldn't be that bad. A lot of the code is about creating the right path objects for the rsa keys, which isn't terribly complicated.

dsmits:
> > I had issues with running unit tests on this script that is in the bin dir so I moved it to the root dir. I could have somehow added the bin dir to the pythonpath but I wasn't sure which approach would've been best. Any suggestions?
> 
> Hmmm, not sure. Was this locally or in a github action? Let's have a look together.

It was in github actions. The reason is that bin is not part of either $PATH or $PYTHONPATH. We could fix it by doing
```
export PYTHONPATH=$PYTHONPATH:$PWD/bin
``
or something, but there are multiple solutions

File: setup_profile.py [hunk 5]
```diff
@@ -0,0 +1,100 @@
...
+        client = NanopubClient()
+        client.publish(np)
+
+
+def _delete_keys():
+    os.remove(DEFAULT_PUBLIC_KEY_PATH)
+    os.remove(DEFAULT_PRIVATE_KEY_PATH)
+
+
+def _declare_this_is_me(orcid: str, public_key: str, name: str) -> Tuple[Graph, BNode]:
```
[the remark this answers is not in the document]
dsmits: I made it `create_this_is_me_rdf` because I like functions to have verb names.

setup_profile.py
```diff
@@ -0,0 +1,100 @@
...
+              prompt='Would you like to publish your profile to the nanopub servers?')
+def main(orcid, publish, name, keypair: Union[Tuple[Path, Path], None]):
+    click.echo('Setting up nanopub profile...')
+
+    if not USER_CONFIG_DIR.exists():
+        USER_CONFIG_DIR.mkdir()
+
+    if not keypair:
+        if _rsa_keys_exist():
+            if _check_erase_existing_keys():
```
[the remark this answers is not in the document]
dsmits: Fixed it

setup_profile.py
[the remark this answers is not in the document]
svenvanderburg: Agree :)

tests/test_setup_profile.py
svenvanderburg: OK, let's keep it!

setup_profile.py
```diff
@@ -0,0 +1,145 @@
...
+                      'this links your ORCID to your RSA key, thereby making all your'
+                      'publications linkable to you'))
+def main(orcid, publish, name, keypair: Union[Tuple[Path, Path], None]):
+    """
+    Interactive CLI to create a user profile.
+
+    :param orcid: the users orcid or other form of universal identifier
+    :param publish: if True, profile will be published to nanopub servers
+    :param name: the name of the user
+    :param keypair: a tuple containing the paths to the public and private RSA key to be used to sign nanopubs. If
```
svenvanderburg, line 48: We use google docstring formatting throughout the rest of the project, I'll change it.

setup.py
```diff
@@ -38,7 +38,7 @@ def get_version(rel_path):
         ],
     },
     include_package_data=True,
-    scripts=['bin/make_nanopub_keys'],
+    scripts=['bin/setup_profile.py'],
```
svenvanderburg, line 41: This is incorrect right?

svenvanderburg: I'm not a big fan of having setup_profile.py in root anyway.

svenvanderburg: I moved `setup_profile.py` to `nanopub/setup_profile.py`.

PR #31, opened 2020-11-04
Use user profile

dsmits:


nanopub/profile.py
```diff
@@ -0,0 +1,28 @@
...
+
+    global _profile
+    if not _profile:
+        _profile = _load_profile()
+    return _profile
+
+
+def _load_profile():
+    path = PROFILE_PATH
+    with path.open('r') as f:
```
svenvanderburg, line 27:
```suggestion
    with  PROFILE_PATH.open('r') as f:
```

nanopub/profile.py
```diff
@@ -0,0 +1,28 @@
+from pathlib import Path
+
+from nanopub.definitions import PROFILE_PATH
+import yaml
+
+_profile = None
+
+
+def get_orcid():
```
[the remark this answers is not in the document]
svenvanderburg: Also, I would expect this to return the id '0000-0000-0000-0000' and not the URI. Maybe add some more documentation so this is clear.

tests/test_nanopub.py
```diff
@@ -9,194 +9,211 @@
...
-
-    nanopub = Nanopub.from_assertion(
-        assertion_rdf=assertion_rdf,
-        uri=rdflib.term.URIRef(test_uri),
-        introduces_concept=rdflib.term.BNode('DrBob'),
-        derived_from=rdflib.term.URIRef('http://www.example.com/another-nanopub')
-        )
-    assert str(nanopub.introduces_concept) == test_concept_uri
+TEST_ASSERTION = (namespaces.AUTHOR.DrBob, namespaces.HYCL.claims, rdflib.Literal('This is a test'))
+TEST_ORCID = rdflib.URIRef('http://example.org/pietje')
```
svenvanderburg, line 13: This doesn't match with what get_orcid returns right?

tests/test_nanopub.py
```diff
@@ -9,194 +9,211 @@
...
+    return mock_profile
 
-    for search in searches:
-        results = client.find_nanopubs_with_text(search)
-        assert len(results) > 0
 
-    assert len(client.find_nanopubs_with_text('')) == 0
+# Created a class so profile can be mocked for all tests at once
+@mock.patch('nanopub.nanopub.profile', _get_mock_profile())
+class NanopubTest(TestCase):
```
svenvanderburg, line 25: Do we need to subclass `TestCase` here? If not maybe leave it out?

svenvanderburg:
```suggestion
class TestNanopub:
```
I would also say this is more common, but curious what you think!

svenvanderburg: Also now that this is in a class, it would be good to make a distinction between tests for Nanopub class and NanopubClient. (i.e, TestNanopub & TestNanopubClient)

tests/test_nanopub.py
svenvanderburg: Also I find the example confusing, because I would expect an ORCID iD like 0000-0000-0000-0000.

tests/test_nanopub.py
```diff
@@ -9,194 +9,211 @@
...
-        assertion_rdf=assertion_rdf,
-        introduces_concept=test_concept,
-    )
-    pubinfo = client.publish(nanopub)
-    assert pubinfo['nanopub_uri'] == test_published_uri
-    assert pubinfo['concept_uri'] == expected_concept_uri
+        with mock.patch('nanopub.nanopub.profile', _get_mock_profile()):
+            result = Nanopub.from_assertion(assertion_rdf=assertion)
+
+            prov_entries = list(
```
svenvanderburg, line 216: I would prefer something like `assert triple in result.rdf`, which would be easier to read.
Also, you could reuse TEST_ORCID.

nanopub/nanopub.py
```diff
@@ -66,6 +67,9 @@ def from_assertion(cls, assertion_rdf, uri=DEFAULT_URI, introduces_concept=None,
 
         """
 
+        if not attributed_to and not nanopub_author:
+            nanopub_author = rdflib.URIRef(profile.get_orcid())
+
```
svenvanderburg, line 72:
```suggestion
        if attributed_to is None:
            attributed_to is rdflib.URIRef(profile.get_orcid())
        if nanopub_author is None:
            nanopub_author = rdflib.URIRef(profile.get_orcid())
```
I think for both we would like to fall back to the ORCID iD.

nanopub/profile.py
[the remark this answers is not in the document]
svenvanderburg: Make sure it also matches what we store in the yaml file, because right now we store the ID, not the URI.

nanopub/nanopub.py
svenvanderburg: The nanopub_author is for the whole nanopub, the attributed_to is for the assertion only.

svenvanderburg (changes requested):
Nice, this is a really cool feature :) I do have some comments. It seems like you didn't test this for realz, you can do that without polluting the nanopub server by using the nanopub test server (`NanopubClient(use_test_server=True)`).

nanopub/profile.py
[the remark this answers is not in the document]
dsmits: > Also, I would expect this to return the id '0000-0000-0000-0000' and not the URI. Maybe add some more documentation so this is clear.

Selection_072

To me it looks like the entire URI is considered the ID

nanopub/profile.py
```diff
@@ -0,0 +1,28 @@
...
+    global _profile
+    if not _profile:
+        _profile = _load_profile()
+    return _profile
+
+
+def _load_profile():
+    path = PROFILE_PATH
+    with path.open('r') as f:
+        return yaml.load(f)
```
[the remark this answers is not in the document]
dsmits: Cool! Didn't really know about that decorator

tests/test_nanopub.py
dsmits: Yeah we need to discuss what is considered the actual ORCID ID

tests/test_nanopub.py
dsmits: yeah if it were up to me the nanopub client would be in a separate module called NanopubClient.
In the `__init__.py` we could then say
```
from NanopubClient import *
```

And then I would put the tests in a separate module as well. What do you think?

tests/test_nanopub.py
```diff
@@ -9,194 +9,211 @@
...
-        (test_concept, namespaces.HYCL.claims, rdflib.Literal('This is a test')))
-
-    nanopub = Nanopub.from_assertion(
-        assertion_rdf=assertion_rdf,
-        introduces_concept=test_concept,
-    )
-    pubinfo = client.publish(nanopub)
-    assert pubinfo['nanopub_uri'] == test_published_uri
-    assert pubinfo['concept_uri'] == expected_concept_uri
+        with mock.patch('nanopub.nanopub.profile', _get_mock_profile()):
```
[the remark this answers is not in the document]
dsmits: because I'm dumb

nanopub/profile.py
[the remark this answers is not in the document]
dsmits: I will use `lru_cache` because `functools.cache` has only been introduced in python 3.8

nanopub/nanopub.py
dsmits: Oh ok so it is actually more reasonable that the user is the author of the nanopub? Then I would vote for just defaulting nanopub_author to the user. We don't know whether the assertion is something the author discovered/made/invented/whatever or that he/she has just found something from a paper and wants to publish it as nanopub

tests/test_nanopub.py
dsmits: By the way, I tried to remove the subclassing stuff but pytest is unable to find the test if I do not subclass from testcase so I left it in. Let me know if you have a solution for that

tests/test_nanopub.py
dsmits: Yeah that would be more readable. The problem is that I do not want to assume what the subject in the triple would look like.
I know it looks like `(SOMETHING, prov:wasAttributedTo, <https://orcid.org/pietje>)`
And from running the test with the debugger I know SOMETHING to be `http://purl.org/nanopub/temp/mynanopub#`

Is it reasonable to assume that it will always be that? Right now the test only checks the values I'm interested in.

About reusing `TEST_ORCID` I am never sure if I should reuse a value that I have also used as input so that is why I copied it. However, I'm never consistent with this is I will just replace it with the `TEST_ORCID`

nanopub/profile.py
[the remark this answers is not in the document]
dsmits: I've gathered more evidence (from https://support.orcid.org/hc/en-us/articles/360006897674-Structure-of-the-ORCID-Identifier):

> ##Format of the ORCID iD
> 
> The ORCID iD is an https URI with a 16-digit number that is compatible with the ISO Standard (ISO 27729), also known as the International Standard Name Identifier (ISNI), e.g. https://orcid.org/0000-0001-2345-6789

dsmits:
Oh damnit, this messes with creating the this-is-me nanopub. I will fix it on monday

nanopub/profile.py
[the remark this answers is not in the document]
svenvanderburg: Sounds good! I was actually looking for the link to `lru_cache` and saw that there was a simpler one. Good to know that it is only introduced in python 3.8.

tests/test_nanopub.py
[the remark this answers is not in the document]
svenvanderburg: Haha, or just lazy.

tests/test_nanopub.py
svenvanderburg: Yeah, I agree to split client and Nanopub class. I will pick it up in a separate PR.

And see my suggestion for renaming `NanopubTest` to `TestNanopub`. Then pytest will be able to find it.

tests/test_nanopub.py
svenvanderburg: See [rdflib documentation](https://rdflib.readthedocs.io/en/stable/intro_to_graphs.html):

Note that this triple does not have to be completely bound:
```python
if (bob, None, None) in graph:
    print("This graph contains triples about Bob!")
```

I think you should expect it to be the value in TEST_ORCID, not caring about what that value is. That makes it also clearer to read (you don't need to worry about whether something actually changed in the string)

nanopub/profile.py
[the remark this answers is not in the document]
svenvanderburg: You are right! Let's make sure we're talking about the url ID by adding some examples in docstrings.

nanopub/profile.py
```diff
@@ -0,0 +1,25 @@
...
+
+import yaml
+
+from nanopub.definitions import PROFILE_PATH
+
+_profile = None
+
+
+def get_orcid_id():
+    return get_profile()['orcid']
```
svenvanderburg, line 12: Let's also store it as orcid_id

tests/test_nanopub.py
svenvanderburg: You are right that the orcid id is the URI + code. `get_orcid` returns a string. So I propose:
```python
TEST_ORCID_ID = 'https://orcid.org/0000-0000-0000-0000'
```

tests/test_nanopub.py
dsmits: it works

tests/test_nanopub.py
dsmits: Nice I changed it

svenvanderburg: Great

nanopub/nanopub.py
svenvanderburg: This is how I see it:

1. By default, the user in the profile is both the nanopub author and the assertion author.
2. If someone else is the nanopub author the user should provide the nanopub_author himself.
3. If someone else is the assertion author the user should provide the attributed_to himself.

The introduction of `use_profile` only complicates things and doesn't add anything IMO. You could argue that you want to protect users from accidentally publishing something attributing to themselves. To prevent that you could indeed have a `use_profile` argument, but it should by default be False and if`use_profile` is False we should raise errors when `attributed_to` argument is empty. Let's discuss!

svenvanderburg: I implemented this.

dsmits (commented):
toppie

PR #32, opened 2020-11-04
Validate ORCID iD

svenvanderburg:


nanopub/setup_profile.py
```diff
@@ -21,6 +22,19 @@
 RSA = 'RSA'
 
 
+def validate_orcid_id(ctx, orcid_id: str):
+    """
+    Check if valid ORCID iD, should be 16 digit in form: 0000-0000-0000-0000
+    """
+    if not orcid_id:
+        return None
+    elif re.match(r'^(\d{4}-){3}\d{4}$', orcid_id):
```
dsmits, line 31: I would create a constant for this regular expression

dsmits (changes requested):
One little suggestion.
Other than that, we need to discuss what is an ORCID ID

svenvanderburg:
I changedf ORCID iD to be in this form: https://orcid.org/1234-5678-1234-5678 instead of just the 16-digit code.

PR #34, opened 2020-11-11
Move NanopubClient to its own module client.py

svenvanderburg:
Me and @dsmits discussed that this would clean up things a bit.


tests/test_client.py
```diff
@@ -0,0 +1,158 @@
...
+
+
+def _get_mock_profile():
+    mock_profile = mock.MagicMock()
+    mock_profile.get_orcid_id.return_value = TEST_ORCID_ID
+
+    return mock_profile
+
+
+@mock.patch('nanopub.nanopub.profile', _get_mock_profile())
```
svenvanderburg, line 23: This is exactly the same as in test_nanopub.py, but I will create a separate PR to fix this in a nice way (by just setting up the profile in the github action)

raar1: OK!

raar1 (approved):
This is indeed much cleaner now.

PR #35, opened 2020-11-11
Setup profile in github actions

svenvanderburg:
* Setup profile in github actions
* Remove mocking of profile from most tests apart from where it is really nessary
* For some reason python was confused that there was a module nanopub/nanopub.py. Then I realised that it is somewhat confusing so I renamed it into nanopub/publication.py and renamed Nanopub to Publication.
* Moved DEFAULT_URI to definitions.py


nanopub/__init__.py
```diff
@@ -1,3 +1,3 @@
-from .nanopub import Nanopub
+from .models import Nanopub
```
svenvanderburg, line 1: I am not completely convinced with the name `models.py`. Any thoughts?

raar1: Hmm, ```nanopub_class.py```?

raar1: ```nanopublication```? We could also change to class name to ```Nanopublication``` to avoid a similar confusion with the name of the library?

nanopub/client.py
```diff
@@ -201,13 +202,13 @@ def publish(self, nanopub: Nanopub):
...
             concept_uri = str(nanopub.introduces_concept)
-            # Replace the DEFAULT_URI with the actually published nanopub uri. # This is
+            # Replace the DEFAULT_NANOPUB_URI with the actually published nanopub uri. # This is
             # necessary if a blank node was passed as introduces_concept. In that case the
             # Nanopub.from_assertion method replaces the blank node with the base nanopub's URI
             # and appends a fragment, given by the 'name' of the blank node. For example, if a
             # blank node with name 'step' was passed as introduces_concept, the concept will be
             # published with a URI that looks like [published nanopub URI]#step.
-            concept_uri = concept_uri.replace(DEFAULT_URI, nanopub_uri)
+            concept_uri = concept_uri.replace(DEFAULT_NANOPUB_URI, nanopub_uri)
```
raar1, line 211: I feel like at some point we should move to using named blank nodes for this stuff as well. In ```from_assertion```, for example, an rdflib blank node is automatically converted to ```NANOPUB_URI#BLANK_NODE_NAME```. Ideally we'd find a way to stop the user setting the nanopub URI at all, since it just leads to confusion because it's a dummy URI... It's more for historical reasons that the user can still set that. Any ideas?

nanopub/models.py
```diff
@@ -32,7 +31,7 @@ def __init__(self, rdf=None, source_uri=None):
                     f'Expected to find {expected} graph in nanopub rdf, but not found. Graphs found: {list(self._graphs.keys())}.')
 
     @classmethod
-    def from_assertion(cls, assertion_rdf, uri=DEFAULT_URI, introduces_concept=None,
+    def from_assertion(cls, assertion_rdf, uri=DEFAULT_NANOPUB_URI, introduces_concept=None,
```
raar1, line 34: This is the bit I was talking about in the earlier comment - allowing ```uri``` to be set here is very confusing to users as it is actually just a dummy URI. I think I had this option originally because I needed a way to convert triples with local URIs to use the final published URI (usually + some fragment). But now this is done with rdflib blank nodes (assigned a name) so I think that we might in fact be able to get rid of this confusing ```uri``` kwarg entirely.

raar1 (approved):
Looks good, left a couple of tangentially related comments as usual :)

nanopub/client.py
svenvanderburg: I've created #38

nanopub/models.py
svenvanderburg: I've created #38

nanopub/__init__.py
svenvanderburg: Yeah I like that. I've renamed the module to `publication` and the class name to `Publication`. (As it is clear it is a nanopublication because of the nanopub namespace). If you like it please merge!

Issue #42, opened 2020-11-13
Allow ```derived_from``` in from_assertion() to take a list


raar1:
We are working with @miguel-ruano on using the ```nanopub``` library in the nice Nanotate tool he and Olga are developing. At present Miguel needs the nanopubs to be able to have multiple ```wasDerivedFrom``` triples in the provenance, due to dependence on a number of annotations. We should extend ```from_assertion()``` to allow ```derived_from``` to (optionally) take a list of URIs, and generate the triples for all of them.

raar1:
Fixed in PR #43

PR #43, opened 2020-11-13
Issue42 allow derived from to take list

raar1:
Addressing issue #42


tests/test_nanopub.py
```diff
@@ -30,6 +30,29 @@ def test_construction_with_bnode_introduced_concept(self):
...
+    def test_construction_with_derived_from_as_list(self):
+        """
+        Test Publication construction from assertion where derived_from is a list.
+        """
+        test_uri = 'http://www.example.com/my-nanopub'
+        derived_from_list = [   'http://www.example.com/another-nanopub', # This nanopub is derived from several sources
+                                'http://www.example.com/and-another-nanopub',
+                                'http://www.example.com/and-one-more' ]
+        assertion_rdf = rdflib.Graph()
+        assertion_rdf.add((rdflib.term.BNode('DrBob'),
```
svenvanderburg, line 42: There's a constant for this assertion on line 8 :)

tests/test_nanopub.py
```diff
@@ -30,6 +30,29 @@ def test_construction_with_bnode_introduced_concept(self):
         )
         assert str(publication.introduces_concept) == test_concept_uri
 
+    def test_construction_with_derived_from_as_list(self):
```
svenvanderburg, line 33: Could you rename this test module to test_publication.py? (I fucked that up)

nanopub/publication.py
```diff
@@ -45,8 +45,8 @@ def from_assertion(cls, assertion_rdf, uri=DEFAULT_NANOPUB_URI, introduces_conce
         derived from the nanopub's URI with a fragment (#) made from the blank node's name.
 
         Args:
-            derived_from: Add that this nanopub prov:wasDerivedFrom the given URI to the provenance
-                graph
+            derived_from: Add that this nanopub prov:wasDerivedFrom the given URI to the provenance graph.
+                          If a list of triples is passed, a provenance triple will be generated for each.
```
svenvanderburg, line 49: A list of strings is passed right? Not triples?

nanopub/publication.py
```diff
@@ -130,12 +130,19 @@ def from_assertion(cls, assertion_rdf, uri=DEFAULT_NANOPUB_URI, introduces_conce
                             attributed_to))
 
         if derived_from:
-            # Convert derived_from URI to an rdflib term first (if necessary)
-            derived_from = rdflib.URIRef(derived_from)
-
-            provenance.add((this_np.assertion,
-                            namespaces.PROV.wasDerivedFrom,
-                            derived_from))
+            list_of_URIs = []
```
svenvanderburg, line 133:
```suggestion
            uris = []
```
Don't use capitals in variable names and it is is obvious that it is a list (since it's plural)

svenvanderburg: And you actually don't have to to declare this variable because you are doing it in the if/else statements.

svenvanderburg (commented):
Nice addition! I've left a few comments.

tests/test_nanopub.py
raar1: Ah, I got this from ```test_construction_with_bnode_introduced_concept()```, should that one also use ```TEST_ASSERTION```?

raar1: Actually ```test_from_assertion()``` is also not using ```TEST_ASSERTION```. I've updated all these test functions to use it.

tests/test_nanopub.py
raar1: Done :)

nanopub/publication.py
raar1: Ah sorry, yep, too tired... A list of strings or ```rdflib.URIRef``` are possible. Fixed now.

nanopub/publication.py
raar1: OK, fair enough. I wanted to avoid confusion with the ```uri``` param, but that will hopefully be removed soon anyway.

Issue #44, opened 2020-11-13
Issue with intro nanopub


raar1:
Using the ```setup_profile``` tool to publish an intro nanopub resulted in the following: [http://server.nanopubs.lod.labs.vu.nl/RAse5_KxBSe4YuEn1B7GWCmJ3DaDxJ1_WGF7gQ81pgilg](http://server.nanopubs.lod.labs.vu.nl/RAse5_KxBSe4YuEn1B7GWCmJ3DaDxJ1_WGF7gQ81pgilg)

Which has the orcid as ```https://orcid.org/https://orcid.org/0000-0002-9984-2720```.

The following was my input to the setup tool:

```
If the public and private key you would like to use are not in /Users/robinrichardson/.nanopub, provide them here. If they are in this directory or you wish to generate new keys, leave empty. [()]:
What is your ORCID iD (i.e. https://orcid.org/0000-0000-0000-0000)? Optionally leave empty []: https://orcid.org/0000-0002-9984-2720
What is your full name?: Robin Richardson
Would you like to publish your profile to the nanopub servers? This links your ORCID iD to your RSA key, thereby making all your publications linkable to you [Y/n]: Y
```

So the provided orcid was in the requested form.

raar1:
Fixed in PR #45

PR #45, opened 2020-11-16
Fix orcid id URI in this-is-me publication

svenvanderburg:


svenvanderburg:
Fixes #44

raar1 (approved):
Oh wow, that was fast ;)

Issue #19, opened 2020-10-26
Improve claim() method


svenvanderburg:
Current `claim()` method is still buggy, it should be a quick way to make a claim. I.e. I (the author in profile, see https://github.com/fair-workflows/nanopub/issues/9), claim 'that tomatoes are red'.

- Change it to only accept a string that is the claim, eg 'tomatoes are red'.
- Make user of the author in the profile, see https://github.com/fair-workflows/nanopub/issues/9

raar1:
I would suggest that we might allow it to be either a string, or alternatively a single RDF triple. We can check the type of what is passed to determine which it is.

At present the ```claim()``` method is a quick way to publish an assertion with a single triple of the form:
```
( AUTHOR, HYCL.claims, Literal(text) )
```

I'm wondering if it would be cleaner to have the assertion itself be the claim and the author doing the claiming be in the provenance subgraph. I guess then the assertion would look more like e.g.:

```
( NPURI#mystatement, RDF.type, HYCL.Statement )
( NPURI#mystatement, RDF.label, Literal(text) )
```

and the provenance graph would contain:

```
( AUTHOR, HYCL.claims, NPURI#mystatement )
```

I'll need to check with @tkuhn to see if this is a correct way of doing this. I'm also not too sure how you would allow a general triple to be the statement, unless you put it in a subgraph, or say the whole assertion is the statement (which might violate the HYCL.Statement type?).

tkuhn:
What is the use case for this? For what will this feature be used?

I am asking because depending on the answer, you could also use AIDA sentences: https://github.com/tkuhn/aida

AIDA statements can then be represented as URI like this one: http://purl.org/aida/Malaria+is+transmitted+by+moquitoes.

This has the benefit that people are encouraged to reuse (i.e. link) existing statements and that identical statements automatically get the same identifier. Like that, you don't need separate labels and can use these statements also in subject position (which you couldn't do with literals).

svenvanderburg:
Nice suggestion @tkuhn!
The use case is to just have a quick one-liner to do statements. It is mostly to show how simple it can be to just publish and not worry about the complex RDF that is created behind the science.

So it will be like:
```python
client = NanopubClient()
client.claim('All people are crazy')
```
Which will publish a nanopub with in the assertion:
( AUTHOR, HYCL.claims, http://purl.org/aida/all+people+are+crazy )

Where AUTHOR is the ORCID iD from profile.

tkuhn:
[...]

Issue #36, opened 2020-11-11
Adding more kwargs to from_assertion() for provenance and pubinfo subgraphs?


raar1:
In principle many useful triples may be desirable in the provenance subgraph. Some of these, like wasDerivedFrom are so common they could well belong as kwargs in their own right, but there should maybe be a way to add other triples to the provenance graph at this stage?

One possibility is to simply have **kwargs in the function signature, and check the keys of that dict for any that match a given list of predicates from the http://www.w3.org/ns/prov# ontology. That would give users the ability to add any of those properties quickly, without our code ballooning up.

We have a similar issue with pubinfo, potentially, but I'm not sure that is such a common one that it necessarily needs to be in from_assertion.

svenvanderburg:
To make this a generic approach we decided to add following arguments:
```python
provenance_triples: List(triples)
pubinfo_triples: List(triples)
```
So a user can add any triple to either the provenance or pubinfo. We will keep the common ones that we already have implemented now. We can always add more kwargs for common properties later.

svenvanderburg:
Also fix this TODO: https://github.com/fair-workflows/nanopub/blob/055631fb13ad5b05ade4f7dfe124de6fb0c23ebf/nanopub/client.py#L244

PR #47, opened 2020-11-19
Improve claim method

svenvanderburg:
See resulting nanopublicaiton: http://test-server.nanopubs.lod.labs.vu.nl/RADSboX94R1mAPIZmAO6VxfNuLr50HYWRmzxonF6F5t9s

### DIFF
--- a/nanopub/client.py
+++ b/nanopub/client.py
@@ -214,7 +214,7 @@ class NanopubClient:
 
         return publication_info
 
-    def claim(self, statement_text: str, author: str = None):
+    def claim(self, statement_text: str):
         """Quickly claim a statement.
 
         Constructs statement triples around the provided text following the Hypotheses and Claims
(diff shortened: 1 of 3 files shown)


nanopub/client.py
```diff
@@ -214,16 +214,35 @@ def publish(self, nanopub: Publication):
 
         return publication_info
 
-    def claim(self, text, rdftriple=None):
-        """
-        Publishes a claim, either as a plain text statement, or as an rdf triple (or both)
-        """
-        assertion_rdf = rdflib.Graph()
+    def claim(self, statement_text: str, author: str = None):
```
svenvanderburg, line 217: I decided to only allow claiming with a statement_text and not with a triple even though that was mentioned in #19 . If you want to publish a triple you can use from_assertion method of Publication. This is just to quickly publish something.

nanopub/client.py
```diff
@@ -214,16 +214,35 @@ def publish(self, nanopub: Publication):
...
+                                 'an ORCID iD (example: https://orcid.org/0000-0000-0000-0000)'
+                                 'for the author argument')
+            else:
+                author = rdflib.URIRef(profile.get_orcid_id())
+        else:
+            author = rdflib.URIRef(author)
+        publication = Publication.from_assertion(assertion_rdf=assertion_rdf,
+                                                 nanopub_author=author,
+                                                 attributed_to=author)
+        # TODO: This is a hacky solution, should be changed once we can add provenance triples to
```
svenvanderburg, line 244: I added that this should be fixed in #36

svenvanderburg:
Fixes #19

raar1 (approved):
All looks good to me. At some point in future maybe we will revisit the idea of AIDA sentences and do something here with them.

Issue #8, opened 2020-10-13
Add some sort of ```retract()``` method


raar1:
I believe you can retract nanopubs that have been published, if you have the original key it was signed with. We will want to have this capability in the client at some stage.

svenvanderburg:
@raar1 Do you have any pointers to documentation on nanopub deletion? I couldn't find anything in the [nanopub-java paper](https://arxiv.org/pdf/1508.04977.pdf), and in the [nanopub guidelines](http://www.nanopub.org/guidelines/1.8/) I find this:
_As the primary purpose of nanopublication is to attach a reputation to a scientific assertion, it will not be possible to delete nanopublications once they have been released. Citation and impact metrics need to consider the entire legacy of one’s nanopublication record and so nanopublications must be preserved in the public record indefinitely. This includes nanopublications that may subsequently be ‘proved’ to be false or in error at a later date. Also, it is important to preserve the legacy of nanopublications in order to reveal research trends that could be exploited for knowledge discovery. With this in mind, nanopublications can be revised only to delete, add or correct information in the Provenance section (attribution, supporting). Any change in the Assertion section constitutes an entirely new nanopublication. Nanopublication revisions must contain meaningful edits, and cannot be exact duplications of existing nanopublications. Earlier versions of nanopublications can be “deprecated” by including dc:isReplacedBy <new_nanopub_uri> . By defualut, OPS will search for and cache only the most recent versions of any nanopulbication._ From which I would conclude that deletion is not possible.

tkuhn:
Deletion is not possible, but nanopubs can be declared retracted. Example of such a retraction nanopub: http://purl.org/np/RAYskLSM5x29icArnWvo9nVrIVEN2mfPoDq3TQSgm-9kk

If the nanopub to be retracted is signed, then the retraction should be signed with the same key.

svenvanderburg:
Thanks for the clarification @tkuhn! I guess this was all already in the mind of @raar1, as it is obvious he is proposing a `retract()` method, and not a `delete()` method... But you made me see the light: I was understanding retraction as something we had to do with the to-be-retracted nanopublication, but instead we retract it by publishing a new nanopublication retracting the to-be-retracted nanopublication.

The example is really useful!

PR #48, opened 2020-11-20
Add retract method

### LINKED ISSUE #46: Agree a fix for the (wrong) URI returned when publishing to test server
I pulled out a snippet from what @miguel-ruano said in an issue on Nanotate:

> i made the update of nanopub in nanotate and make changes according of the new version and works perfect!!!,
> thanks for the update,
> 
> testing in test mode, i see something when publishing the nanopublication the return url is like http://purl.org/np/RAdG9VXVOBq-7Ju5H085CgynfrVlwpajLb02B-W0ss8Zc,
> when i go to url this returns 404(not found), because the nanopub not is publish in http://server.nanopubs.lod.labs.vu.nl/ if not in http://test-server.nanopubs.lod.labs.vu.nl/
> 
> seeing the nanopub code exactly in the NanopubClient publish method
> 
> the process is the following
> first sign the rdf and then publish it
> 
> in the process of sign the rdf, the code seems not know about the test enviroment and always url base is like http://purl.org/np/RAdG9VXVOBq-7Ju5H085CgynfrVlwpajLb02B-W0ss8Zc
> 
> ```
> def sign(self, unsigned_file: Union[str, Path]) -> str:
> 	unsigned_file = str(unsigned_file)
>         self._run_command(f'{NANOPUB_SCRIPT} sign ' + unsigned_file)
>         return self._get_signed_file(unsigned_file)
> ```
> 
> so when extract the nanopub uri, after publishing in test server
> 
> ```
>  @staticmethod
>  def extract_nanopub_url(signed: Union[str, Path]):
>       # Extract nanopub URL
>       # (this is pretty horrible, switch to python version as soon as it is ready)
>       extracturl = rdflib.Graph()
>       extracturl.parse(str(signed), format="trig")
>       return dict(extracturl.namespaces())['this'].__str__()
> ```
> 
> always return the same url for test or production mode, i guess that bug is in the sign method.
> 
> I hope this helps you to solve it in the future.

Can we make a quick fix for this by just modifying the wrapper methods he mentions above? Or is there something else we need to be careful about?

svenvanderburg:
* Add retract method
* When the nanopublication is signed with a public key that does not the public key in the profile we raise an error. This behavior can be overridden with `force` argument.
* For this to work using the test server I updated the fetch method to work with test-server publications. That should fix #46 too.


svenvanderburg:
See [resulting retraction nanopublication](http://test-server.nanopubs.lod.labs.vu.nl/RAypzNlnRli8LnHGCz-O2m3phCbuFX3Z5fDq_XaClKSoI)

svenvanderburg:
Fixes #8

nanopub/client.py
```diff
@@ -172,14 +171,15 @@ def fetch(uri, format: str = 'trig'):
         else:
             raise ValueError(f'Format not supported: {format}, choose from '
                              f'{[format.value for format in Formats]})')
-
+        if self.use_test_server:
```
svenvanderburg, line 174: This has the implication that whenever a user creates a NanopubClient(use_test_server=True) you can only fetch test nanopublicaiton. I kind of like that, you are sort of in the test environment, but I'm curious what the reviewer thinks.

tests/test_client.py
```diff
@@ -88,15 +88,14 @@ def test_nanopub_fetch(self):
         Check that Nanopub fetch is returning results for a few known nanopub URIs.
         """
         known_nps = [
-            'http://purl.org/np/RAFNR1VMQC0AUhjcX2yf94aXmG1uIhteGXpq12Of88l78',
-            'http://purl.org/np/RAePO1Fi2Wp1ARk2XfOnTTwtTkAX1FBU3XuCwq7ng0jIo',
-            'http://purl.org/np/RA48Iprh_kQvb602TR0ammkR6LQsYHZ8pyZqZTPQIl17s'
+            'http://purl.org/np/RANGY8fx_EYVeZzJOinH9FoY-WrQBerKKUy2J9RCDWH6U',
+            'http://purl.org/np/RAABh3eQwmkdflVp50zYavHUK0NgZE2g2ewS2j4Ur6FHI',
+            'http://purl.org/np/RA8to60YFWSVCh2n_iyHZ2yiYEt-hX_DdqbWa5yI9r-gI'
```
svenvanderburg, line 93: These now point to test server URIs

nanopub/client.py
```diff
@@ -174,12 +173,16 @@ def fetch(uri, format: str = 'trig'):
                              f'{[format.value for format in Formats]})')
 
         r = requests.get(uri + extension)
+        if not r.ok and self.use_test_server:
```
raar1, line 176: This is good. I like that it does check the real servers first.

nanopub/client.py
```diff
@@ -227,3 +230,48 @@ def claim(self, text, rdftriple=None):
...
+        """ Check for matching public keys of a nanopublication with the profile.
+
+        Raises:
+            AssertionError: When the nanopublication is signed with a public key that does not
+                match the public key in the profile
+        """
+        publication = self.fetch(uri)
+        their_public_keys = list(publication.pubinfo.objects(rdflib.URIRef(uri + '#sig'),
+                                                             namespaces.NPX.hasPublicKey))
+        if len(their_public_keys) > 0:
```
raar1, line 244: Can this list in principle be more than length 1? Should we check and warn if that is the case?

tests/test_client.py
raar1: Nice

tests/test_client.py
```diff
@@ -146,3 +145,33 @@ def test_nanopub_publish_blanknode(self):
         pubinfo = client.publish(nanopub)
         assert pubinfo['nanopub_uri'] == test_published_uri
         assert pubinfo['concept_uri'] == expected_concept_uri
+
+    def test_retract_with_force(self):
+        client = NanopubClient()
+        client.java_wrapper.publish = mock.MagicMock()
+        client.retract('http://www.example.com/my-nanopub', force=True)
```
raar1, line 152: I wonder if this is a "true" retraction in the eyes of Tobias, since you're supposed to have exactly the same public key to be able to retract a nanopub. I do agree that having a ```force=True``` option makes sense though.

raar1 (approved):
Good feature to have! I think there should be a short entry for this in the README quickstart too.

nanopub/client.py
svenvanderburg: Yes, it's rdf so in principle anything can be the case...

tests/test_client.py
svenvanderburg: Yeah, good point...

nanopub/client.py
svenvanderburg: I added a warning, but still taking the first in the list.

PR #49, opened 2020-11-20
Remove specifying of format when fetching nanopubs

### LINKED ISSUE #28: Decide what ```Format``` should be - only trig at present
This:

```python
@unique
class Formats(Enum):
    """
    Enums to specify the format of nanopub desired
    """
    TRIG = 'trig'
```

and this:

```python
def fetch(uri, format: str = 'trig'):
```

are starting to look a bit embarassing, and have been hanging around in the code base for ages.

* Check which rdf formats can be returned by general nanopub servers
* Do some servers only serve a subset of these? Does ```fetch()``` need to check if a given format is available first?
* Add more RDF formats to ```Format``` enums, or find a better solution

Alternatively, don't bother with format selection - it gets parsed into an rdflib graph immediately anyway, so the format that is fetched doesn't really matter. But in such a case this code should be cleaned up.

Methods for providing the nanopub rdf serialized into a number of optional formats (trig, ttl, nt etc) would be useful, regardless.

svenvanderburg:
Fixes #28


raar1 (approved):
Simple change, but makes the code base a lot cleaner.

Issue #39, opened 2020-11-13
Enforce profile creation with orcid id


svenvanderburg:
- [x] Get rid of `nanopub_author`, always use profile. If people want to have a different orcid id as nanopub author they should rerun setup_profile.
- [x] If there's no orcid id in profile or no profile raise error and point to setup_profile
- [x] Rename `attributed_to` to `assertion_attributed_to` 
- [x] Rename`attribute_to_profile` to `attribute_assertion_to_profile`

PR #51, opened 2020-11-23
Enforce using profile

svenvanderburg:
- [x] Get rid of `nanopub_author`, always use profile. If people want to have a different orcid id as nanopub author they should rerun setup_profile.
- [x] If there's no orcid id in profile or no profile raise error and point to setup_profile
- [x] Enforce setting a valid orcid_id in setup_profile
- [x] Rename `attributed_to` to `assertion_attributed_to` 
- [x] Rename`attribute_to_profile` to `attribute_assertion_to_profile`

### DIFF
--- a/nanopub/publication.py
+++ b/nanopub/publication.py
@@ -33,7 +33,7 @@ class Publication:
     @classmethod
     def from_assertion(cls, assertion_rdf, uri=DEFAULT_NANOPUB_URI, introduces_concept=None,
                        derived_from=None, attributed_to=None,
-                       attribute_to_profile: bool = False, nanopub_author=None):
+                       attribute_to_profile: bool = False):
         """
         Construct Nanopub object based on given assertion, with given assertion and (defrag'd) URI.
         Any blank nodes in the rdf graph are replaced with the nanopub's URI, with the blank node name
@@ -50,15 +50,8 @@ class Publication:
             attributed_to: the provenance graph will note that this nanopub prov:wasAttributedTo
                 the given URI.
             attribute_to_profile: Attribute the nanopub to the ORCID iD in the profile
-            nanopub_author: the pubinfo graph will note that this nanopub prov:wasAttributedTo the
-                given URI. If no nanopub_author is provided we default to the author from the
-                profile
 
         """
-
-        if nanopub_author is None and profile.get_orcid_id() is not None:
-            nanopub_author = rdflib.URIRef(profile.get_orcid_id())
-
         if attributed_to and attribute_to_profile:
             raise ValueError('If you pass a URI for the attributed_to argument, you cannot pass '
                              'attribute_to_profile=True, because the nanopub will already be '
@@ -135,7 +128,7 @@ class Publication:
                 list_of_URIs = derived_from
             else:
                 list_of_URIs = [derived_from]
-           
+
             for derived_from_uri in list_of_URIs:
                 # Convert uri to an rdflib term first (if necessary)
                 derived_from_uri = rdflib.URIRef(derived_from_uri)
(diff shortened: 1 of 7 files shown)


svenvanderburg:
Fixes #39

nanopub/publication.py
```diff
@@ -32,8 +32,8 @@ def __init__(self, rdf=None, source_uri=None):
 
     @classmethod
     def from_assertion(cls, assertion_rdf, uri=DEFAULT_NANOPUB_URI, introduces_concept=None,
```
raar1, line 34: The code in this method is now a lot clearer (to me)! I'm torn between the simplicity of having ```attribute_to_profile```/```attribute_assertion_to_profile``` be bools, or reducing the number of arguments by having e.g. ```attribute_assertion_to_profile``` default to ```True``` in the case that ```assertion_attributed_to``` remains ```None```. I guess the current way is better because someone might (in principle) not want to attribute the assertion to anyone?

nanopub/setup_profile.py
```diff
@@ -26,9 +26,7 @@ def validate_orcid_id(ctx, orcid_id: str):
     Check if valid ORCID iD, should be https://orcid.org/ + 16 digit in form:
         https://orcid.org/0000-0000-0000-0000
     """
-    if not orcid_id:
-        return None
-    elif re.match(ORCID_ID_REGEX, orcid_id):
+    if re.match(ORCID_ID_REGEX, orcid_id):
         return orcid_id
     else:
         raise ValueError('Your ORCID iD is not valid, please provide a valid ORCID iD that '
```
raar1, line 32: I think this error should reiterate that it must look like ```https://orcid.org/0000-0000-0000-0000```. Even though it does indeed say that in the prompt, people will no doubt run this non-interactively, and if you don't see the help message for ```--orcid_id``` then it would be reasonable to input ```0000-0000-0000-0000``` or some other variation.

tests/test_setup_profile.py
```diff
@@ -55,9 +55,8 @@ def test_provided_keypair_copied_to_nanopub_dir(tmp_path: Path):
...
 
 def test_validate_orcid_id():
-    valid_ids = ['https://orcid.org/1234-5678-1234-5678', '']
-    for orcid_id in valid_ids:
-        validate_orcid_id(ctx=None, orcid_id=orcid_id)
+    valid_id = 'https://orcid.org/1234-5678-1234-5678'
+    assert validate_orcid_id(ctx=None, orcid_id=valid_id) == valid_id
     invalid_ids = ['https://orcid.org/abcd-efgh-abcd-efgh',
                    'https://orcid.org/1234-5678-1234-567',
                    'https://orcid.org/1234-5678-1234-56789',
```
raar1, line 62: Very minor suggestion - maybe add an invalid ID that isn't a URI too (such as one of the form ```0000-0000-0000-0000```) because that is something people might conceivably enter.

raar1 (approved):
Good PR. Usage of library seems more sane now that users are obliged to explicitly set up who is authoring the nanopubs before publishing.

nanopub/publication.py
svenvanderburg: Yes, that is why me and Djura chose for this solution. Because indeed someone might not want to attribute assertion to anyone, but more importantly if we attribute to the profile by default someone might accidentally publish something that is attributed to themselves without them knowing!

tests/test_setup_profile.py
svenvanderburg: Good one, I've added it!

nanopub/setup_profile.py
svenvanderburg: It actually says so in the next line: https://github.com/fair-workflows/nanopub/blob/3d43ca79a35dcddcb23076bed7c6303266d8dc75/nanopub/setup_profile.py#L33

PR #52, opened 2020-11-23
Remove uri arg from assertion

### LINKED ISSUE #38: Remove uri argument from Nanopub.from_assertion

svenvanderburg:
Fixes #38. 
Also includes some refactoring of test_publication.py

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,7 +116,7 @@ replace_in_rdf(rdf=my_assertion, oldvalue=URIRef('www.example.org/timbernerslee'
 ``` 
 
 ### Specifying derived_from
-You can specify that the nanopub is derived from another nanopublication:
+You can specify that the nanopub's assertion is derived from another URI (such as an existing nanopublication):
 ```python
 publication = Publication.from_assertion(assertion_rdf=my_assertion,
                                          derived_from=rdflib.URIRef('www.example.org/another-nanopublication'))
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ You can specify that the nanopub's assertion is derived from another URI (such a
 publication = Publication.from_assertion(assertion_rdf=my_assertion,
                                          derived_from=rdflib.URIRef('www.example.org/another-nanopublication'))
 ```
-,
+Note that ```derived_from``` may also be passed a list of URIs.
                                          
 
 
--- a/README.md
+++ b/README.md
@@ -93,13 +93,35 @@ for s, p, o in np.provenance:
 print(np.introduces_concept)
 ```
 
-### Specifying more information
-You can optionally specify that the ```Publication``` introduces a particular concept, or is derived from another nanopublication:
+### Specifying concepts relative to the nanopublication namespace
+You can optionally specify that the ```Publication``` introduces a particular concept using blank nodes. 
+The pubinfo graph will note that this nanopub npx:introduces the concept. The concept should be a blank node 
+(rdflib.term.BNode), and is converted to a URI derived from the nanopub's URI with a fragment (#) made from the blank
+node's name.
+```python
+from rdflib.term import BNode
+publication = Publication.from_assertion(assertion_rdf=my_assertion,
+                                         introduces_concept=BNode('timbernerslee'))
+```
+Upon publication, any blank nodes in the rdf graph are replaced with the nanopub's URI, with the blank node name as a
+fragment. For example, if the blank node is called 'step', that would result in a URI composed of the nanopub's (base)
+URI, followed by #step. In case you are basing your publication on rdf that has a lot of concepts specific to this 
+nanopublication that are not blank nodes you could use `replace_in_rdf` to easily replace them with blank nodes:
+```python
+from nanopub import replace_in_rdf
+replace_in_rdf(rdf=my_assertion, oldvalue=URIRef('www.example.org/timbernerslee'), newvalue=BNode('timbernerslee'))
+``` 
+
+### Specifying derived_from
+You can specify that the nanopub is derived from another nanopublication:
 ```python
 publication = Publication.from_assertion(assertion_rdf=my_assertion,
-                                         introduces_concept=(URIRef('www.example.org/timbernerslee'),
-                                         derived_from=URIRef('www.example.org/another-nanopublication') )
+                                         derived_from=rdflib.URIRef('www.example.org/another-nanopublication'))
 ```
+,
+                                         
+
+
 
 ## Dependencies
 The ```nanopub``` library currently uses the [```nanopub-java```](https://github.com/Nanopublication/nanopub-java) tool for signing and publishing new nanopublications. This is automatically installed by the library.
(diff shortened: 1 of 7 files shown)


nanopub/publication.py
```diff
@@ -68,32 +94,16 @@ def from_assertion(cls, assertion_rdf, uri=DEFAULT_NANOPUB_URI, introduces_conce
...
-            if isinstance(s, rdflib.term.BNode):
-                s = this_np[str(s)]
-            if isinstance(o, rdflib.term.BNode):
-                o = this_np[str(o)]
-            assertion_rdf.add((s, p, o))
+        if introduces_concept and not isinstance(introduces_concept, rdflib.term.BNode):
+            raise ValueError('If you want a nanopublication to introduce a concept, you need to '
+                             'pass it as an rdflib.term.BNode("concept_name"). This will make '
+                             'sure it is referred to from the nanopublication uri namespace upon '
+                             'publishing.')
```
svenvanderburg, line 102: I think this makes sense now, but what do you think?

tests/test_publication.py
```diff
@@ -1,87 +1,62 @@
...
 from rdflib.namespace import RDF
 
 from nanopub import namespaces, Publication
+from nanopub.definitions import DUMMY_NANOPUB_URI
 
-TEST_ASSERTION = (namespaces.AUTHOR.DrBob, namespaces.HYCL.claims, rdflib.Literal('This is a test'))
 TEST_ORCID_ID = 'https://orcid.org/0000-0000-0000-0000'
 
 
 class TestPublication:
```
svenvanderburg, line 13: I've removed a lot of redundant tests, and made each test test a specific thing.

README.md
```diff
@@ -96,13 +96,35 @@ for s, p, o in np.provenance:
...
-                                         derived_from=URIRef('www.example.org/another-nanopublication') )
+                                         introduces_concept=BNode('timbernerslee'))
 ```
+Upon publication, any blank nodes in the rdf graph are replaced with the nanopub's URI, with the blank node name as a
+fragment. For example, if the blank node is called 'step', that would result in a URI composed of the nanopub's (base)
+URI, followed by #step. In case you are basing your publication on rdf that has a lot of concepts specific to this 
+nanopublication that are not blank nodes you could use `replace_in_rdf` to easily replace them with blank nodes:
+```python
+from nanopub import replace_in_rdf
+replace_in_rdf(rdf=my_assertion, oldvalue=URIRef('www.example.org/timbernerslee'), newvalue=BNode('timbernerslee'))
```
raar1, line 115: This is fine, but I'm wondering if we should instead return a new rdf graph instead of modifying this one in-place? I suppose users might not always want to modify their original graph. Then again, this is pretty minor, so we should just keep it like this unless/until a specific use case shows up. The documentation in the README looks good to me though.

README.md
```diff
@@ -96,13 +96,35 @@ for s, p, o in np.provenance:
...
+``` 
+
+### Specifying derived_from
+You can specify that the nanopub is derived from another nanopublication:
+```python
+publication = Publication.from_assertion(assertion_rdf=my_assertion,
+                                         derived_from=rdflib.URIRef('www.example.org/another-nanopublication'))
+```
+,
```
raar1, line 124:
```suggestion
Note that ```derived_from``` may also be passed a list of URIs.
```

README.md
```diff
@@ -96,13 +96,35 @@ for s, p, o in np.provenance:
...
+fragment. For example, if the blank node is called 'step', that would result in a URI composed of the nanopub's (base)
+URI, followed by #step. In case you are basing your publication on rdf that has a lot of concepts specific to this 
+nanopublication that are not blank nodes you could use `replace_in_rdf` to easily replace them with blank nodes:
+```python
+from nanopub import replace_in_rdf
+replace_in_rdf(rdf=my_assertion, oldvalue=URIRef('www.example.org/timbernerslee'), newvalue=BNode('timbernerslee'))
+``` 
+
+### Specifying derived_from
+You can specify that the nanopub is derived from another nanopublication:
```
raar1, line 119:
```suggestion
You can specify that the nanopub's assertion is derived from another URI (such as an existing nanopublication):
```

nanopub/definitions.py
```diff
@@ -7,5 +7,5 @@
 USER_CONFIG_DIR = Path.home() / '.nanopub'
 PROFILE_PATH = USER_CONFIG_DIR / 'profile.yml'
 
-# Default URI when referring to a nanopub, will be replaced with published URI when publishing.
-DEFAULT_NANOPUB_URI = 'http://purl.org/nanopub/temp/mynanopub'
+# Dummy URI when referring to a nanopub, will be replaced with published URI when publishing.
+DUMMY_NANOPUB_URI = 'http://purl.org/nanopub/temp/mynanopub'
```
raar1, line 11: This is definitely a lot clearer now.

nanopub/publication.py
```diff
@@ -30,21 +31,48 @@ def __init__(self, rdf=None, source_uri=None):
...
+        the blank node is called 'step', that would result in a URI composed of the nanopub's (base)
+        URI, followed by #step.
 
         Args:
+            assertion_rdf: The assertion RDF graph.
+            introduces_concept: the pubinfo graph will note that this nanopub npx:introduces the
+                concept. The concept should be a blank node (rdflib.term.BNode), and is converted
+                to a URI derived from the nanopub's URI with a fragment (#) made from the blank
+                node's name.
             derived_from: Add that this nanopub prov:wasDerivedFrom the given URI to the provenance graph.
```
raar1, line 76:
```suggestion
            derived_from: Add a triple to the provenance graph stating that this nanopub's assertion prov:wasDerivedFrom the given URI.
```

nanopub/publication.py
raar1: I think so. I'm trying to think of any situation in which a nanopub would introduce a concept that is not contained within its assertion subgraph (in some form), in which case you'd obviously have to allow general URIs to be passed. But I don't think that would really ever make sense, so this looks good as is.

I suppose I'd been thinking we could also just accept strings here, but since it would always need to refer to a concept inside the assertion (and since that would need to use ```BNode```) then it is probably best to insist on the user providing a ```BNode``` here. This will save headaches later when they inevitably enter a URI in the string and you get that full URI as a fragment or something horrible like that. I think your error message is clear enough here about what they should do. We will add more to the full documentation once we have it.

tests/test_publication.py
raar1: 👍

raar1 (approved):
Looks great. I've left a couple of comments and suggestions.

README.md
svenvanderburg: Yeah, that's a good point. I am neutral. The user can always do a copy themselves. So I will keep it like this for now.

nanopub/publication.py
svenvanderburg: Nice thoughts. I also thought about passing strings, but indeed then we cannot really check whether it is actually a uri or something else. So good point!

PR #54, opened 2020-11-24
Fix bug public key

svenvanderburg:


nanopub/client.py
```diff
@@ -226,7 +226,7 @@ def claim(self, statement_text: str):
         publication.provenance.add((rdflib.URIRef(profile.get_orcid_id()),
                                     namespaces.HYCL.claims,
                                     rdflib.URIRef(DUMMY_NANOPUB_URI + '#mystatement')))
-        self.publish(publication)
+        return self.publish(publication)
```
svenvanderburg, line 229: This is also useful!

raar1: Agreed.

raar1: Needed for consistency

raar1 (approved):
Simple PR, useful changes

PR #55, opened 2020-11-24
Add notebook

svenvanderburg:
* Add notebook for demoing claim and retract
* Add notebook for demoing blank nodes


raar1 (approved):
Thanks - looks good! This notebook will be useful for the weekly fairworkflows call :)

PR #57, opened 2020-11-25
Optionally provide pubinfo_rdf or provenance_rdf

### LINKED ISSUE #36: Adding more kwargs to from_assertion() for provenance and pubinfo subgraphs?

svenvanderburg:
Fixes #36

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,9 +122,22 @@ publication = Publication.from_assertion(assertion_rdf=my_assertion,
                                          derived_from=rdflib.URIRef('www.example.org/another-nanopublication'))
 ```
 Note that ```derived_from``` may also be passed a list of URIs.
+                               
+### Specifying custom publication info or provenance triples
+You can add your own triples to the provenance graph of the nanopublication:
+```python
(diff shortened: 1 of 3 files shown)


raar1 (approved):
No problems here, and agree totally with the variable name changes. One minor point would be maybe you could add something about these two new params to the Quick Start?

svenvanderburg:
I've added something about it in the Quick Start, although I think once we have user documentation we should move it there (As this is a bit more advanced than quick start).

PR #58, opened 2020-11-25
Use YAtiML to load and store profiles

LourensVeen:
The profile loading/saving code generates a warning about an insecure loader, which doesn't look so good even if you probably trust the input most of the time. After all, there's a link to a private key in there. Also, it doesn't validate input sufficiently, doesn't generate specific error messages, and looks a bit clunky. These things are actually surprisingly hard to do, so there's no shame in that, but did you know that there's a nice library called YAtiML that can do the hard work for you? All you do is create a class that matches your file format, and YAtiML does the rest.

This pull request uses YAtiML to load and store profiles, and adds tests that verify that this works correctly.


nanopub/profile.py
```diff
@@ -20,46 +14,86 @@ class ProfileError(RuntimeError):
...
+        name: The user's name
+        public_key: Path to the user's public key
+        private_key: Path to the user's private key
+        nanopub_uri: URI of the user's profile nanopub
+    """
+    def __init__(
+            self,
+            orcid_id: str, name: str,
+            public_key: Path, private_key: Path,
+            nanopub_uri: Optional[str] = None
```
LourensVeen, line 31: `Profile.profile_nanopub` sounded silly, so I'm calling it `nanopub_uri` on the Python side.

nanopub/profile.py
```diff
@@ -20,46 +14,86 @@ class ProfileError(RuntimeError):
...
+        self.private_key = private_key
+        self.nanopub_uri = nanopub_uri
+
+    @classmethod
+    def _yatiml_savorize(cls, node: yatiml.Node) -> None:
+        node.rename_attribute('profile_nanopub', 'nanopub_uri')
+
+    @classmethod
+    def _yatiml_sweeten(cls, node: yatiml.Node) -> None:
+        node.remove_attributes_with_default_values(cls)
```
LourensVeen, line 46: This removes `nanopub_uri` (or `profile_nanopub`) from the YAML file if it is None, which is what the original did as well.

nanopub/profile.py
svenvanderburg: I think we should rename this everywhere while we're at it. Maybe into `introduction_nanopub_uri` to make it even clearer.

nanopub/profile.py
```diff
@@ -20,46 +14,86 @@ class ProfileError(RuntimeError):
...
+    Raises:
+        yatiml.RecognitionError: If there is an error in the file.
     """
-    path = PROFILE_PATH
-    with path.open('r') as f:
-        return yaml.load(f)
+    try:
+        return _load_profile(PROFILE_PATH)
+    except yatiml.RecognitionError as e:
+        raise ProfileError(str(e))
```
svenvanderburg, line 81: I think we should tell users that their profile was not setup correctly and point them to the documentation.

nanopub/profile.py
```diff
@@ -20,46 +14,86 @@ class ProfileError(RuntimeError):
...
-    return public_key
+
+def get_orcid_id() -> str:
+    """Returns the user's ORCID."""
+    return get_profile().orcid_id
+
+
+def get_public_key() -> str:
+    """Returns the user's public key."""
+    return get_profile().public_key
```
svenvanderburg, line 63: This doesn't return the key itself anymore, but the path to it right? And there's apparently no test that breaks :O

nanopub/profile.py
svenvanderburg: I think to simplify things we don't have to do that. Just leave it empty

tests/test_profile.py
```diff
@@ -1,11 +1,68 @@
...
+
+        with test_file.open('r') as f:
+            assert f.read() == (
+                    'orcid_id: pietje\n'
+                    'name: https://orcid.org/0000-0000-0000-0000\n'
+                    'public_key: /home/.nanopub/id_rsa.pub\n'
+                    'private_key: /home/.nanopub/id_rsa\n')
+
+
+def test_nanopub_uri_roundtrip(tmpdir):
```
svenvanderburg, line 44: What does this test exactly? How is it different than `test_load_profile` and `test_store_profile`?

svenvanderburg:
I wanted to just leave some comments, but it turned out to be a sort of review... I am not sure if you were done though.

I have to say that I find using `yatiml` for this a bit overkill, I think it makes the code unnecessarily complex. Using `yaml.safe_load()` would have already solved half the problem. But we also get validation for free now which is nice. So overall, I approve 👍

nanopub/profile.py
LourensVeen: Oh you're right, I overlooked that, will fix. It passes the tests as-is, I ran them locally before committing of course. So this is a coverage issue, not a CI issue.

tests/test_profile.py
LourensVeen: It's a little belt-and-suspenders, but it has `nanopub_uri` specified, which the other tests don't have.

nanopub/profile.py
LourensVeen: What would the documentation suggest to do? Regenerate the profile? That would overwrite the user's keys and leave them e.g. unable to retract their previous nanopubs. So presumably some kind of manual fix would be needed, in which case it would be useful to know where the problem is exactly.

LourensVeen:
My apologies, there was definitely a wink-and-nudge intended in the initial comment but I'm not sure it came across. I am obviously biased :-).

As for complexity, I guess it is in the eye of the beholder. I think having a single definition of the format is simpler, and having a well-defined class is easier to understand than a dict which could contain anything. So I prefer the YAtiML approach.

If you have suggestions for how I can better communicate the quality, stability and maturity of YAtiML then I'm all ears. It's a lot better than you suggest at least.

nanopub/profile.py
LourensVeen: Well, you can't change the YAML format I think? Unless you want to modify the Java code as well?

nanopub/profile.py
LourensVeen: True, and it would save a line of code...

nanopub/profile.py
LourensVeen: I've added some text to that effect, but also left the detailed error message so users have the information needed to fix things.

nanopub/profile.py
LourensVeen: Fixed, although I'm checking for a missing file now rather than an empty file. I'm not sure what the expected error modes are, but at least you get a missing file if the profile has not been set up.

nanopub/profile.py
svenvanderburg: No, the yaml format is ours! The java code has nothing to do with this. So we can change anything! We could even change `profile_nanopub` to `lourens_is_awesome`.

nanopub/profile.py
svenvanderburg: Yeah, regenerate the profile. Overwriting the user key is optional. The solution you have now is perfect I think (both detailed and user friendly)!

nanopub/profile.py
svenvanderburg: Ha, here's a good argument for increasing test coverage :)

tests/test_profile.py
svenvanderburg: Ok, that seems overkill to me, but then again in the other comment I was saying we need more test coverage.... ;)

LourensVeen: Actually, there should probably be some negative tests which test various broken files and that they're handled correctly. The original code would raise a `KeyError` if the `public_key` attribute was missing, and a `FileNotFoundError` if the whole file was missing. It explicitly handles an empty file, by raising `ProfileError`. So ideally these would all be handled correctly (for whatever is decided to be correct), and tests would exist that would check that.

svenvanderburg: Yeah, fair enough. I think we can be a bit less ambitious, but feel free to add more tests if you want to 😛

nanopub/profile.py
LourensVeen: Changed to `introduction_nanopub_uri` now, and removed the mapping code so that the YAML format now matches the Python class exactly.

nanopub/profile.py
LourensVeen: I removed seasoning completely as discussed.

tests/test_profile.py
LourensVeen: I'll leave it at this then.

nanopub/profile.py
LourensVeen: Actually, if it's all the same to you, maybe `orcid_id` should be `orcid`? It's not the id of the id, but just the id itself.

LourensVeen:
Okay, except for the `orcid_id` -> `orcid` I think everything is done. I'd like to make that change, but let me know what you think.

nanopub/profile.py
svenvanderburg: Nice work! We had quite some dicussions about that, we settled on orcid_id. ORCID refers to it as ORCID iD, see https://support.orcid.org/hc/en-us/articles/360006897334-What-is-my-ORCID-iD-and-how-should-I-use-it- . But in fact ORCID stands for Open Researcher and Contributor ID so they refer to it themselves as Open Researcher and Contributor ID ID.

svenvanderburg:
Nice changes! I am merging :) (See my comment on `orcid` vs `orcid_id` above.)

nanopub/profile.py
LourensVeen: Ha! Well, when in Rome...

PR #66, opened 2020-11-30
Detailed pages for user documentation

svenvanderburg:
Adds documentation for:
* setup
* publishing + retraction
* searching
* fetching

I also added notebooks to the `examples` folder that mostly contain the same examples as in the documentation. I don't think we should really make those mimic each other, documentation is leading. But it is nice for actually testing the code in the documentation.

I suggest reviewing this by building it locally using
```
cd docs
make html
```
Then look at _build/index.html in your browser. (This should go in developer/contributer guide)


README.md
```diff
@@ -65,86 +65,20 @@ from nanopub import NanopubClient
 # Search for all nanopublications containing the text 'fair'
```
svenvanderburg, line 65: I removed all this from README.md so it only includes very basic quickstart.

docs/publishing/publishing-nanopublications.md
```diff
@@ -0,0 +1,128 @@
...
+
+>>> # Create the client (we use use_test_server=True to point to the test server)
+>>> client = NanopubClient(use_test_server=True)
+
+>>> # Publish a simple statement to the server
+>>> client.claim('All cats are gray')
+Published to http://purl.org/np/RAcTwW_JdTqZJTXA4jUTFUj6ut8OI9AqaPb1IBPsLsy-o
+```
+You can [view](http://test-server.nanopubs.lod.labs.vu.nl/RAcTwW_JdTqZJTXA4jUTFUj6ut8OI9AqaPb1IBPsLsy-o)
+the resulting nanopublication on the nanopub test server. (NB: the purl link actually
```
raar1, line 23: I think it is a good idea that all the docs example code uses the test server by default, as it will help reduce pollution when new users try out the library by just copy-pasting code. But I do think it's a little confusing that the user is immediately faced with the peculiarities of the returned URIs not being the real URIs etc. Perhaps we want a short section on test-server vs no test-server before these code snippets come up?

docs/publishing/retraction.md
```diff
@@ -0,0 +1,24 @@
...
+This is done by publishing a new nanopublication that states that you
+retract the original publication. You can use `NanopubClient.retract()`:
+```python
+>>> from nanopub import NanopubClient
+>>> # Create the client (we use use_test_server=True to point to the test server)
+>>> client = NanopubClient(use_test_server=True)
+>>> client.retract('http://purl.org/np/RAcTwW_JdTqZJTXA4jUTFUj6ut8OI9AqaPb1IBPsLsy-o')
+Published to http://purl.org/np/RAOLg94DSpgPVifZZvhwkJ6NBl558pVXaclrgEYhLpfoI
+```
+View the full retraction nanopublication [here](http://test-server.nanopubs.lod.labs.vu.nl/RAOLg94DSpgPVifZZvhwkJ6NBl558pVXaclrgEYhLpfoI).
```
raar1, line 13: Note that the test server is liable to be cleared every so often, so you can't rely on test URIs to still be there in future. We may wish to also publish this test retraction to the real servers if you want to point to it here.

docs/setup.md
```diff
@@ -0,0 +1,44 @@
...
+### Introductory nanopublication
+We encourage you to make use of `setup_profile`'s option 
+to publish your profile to the nanopub servers. This links your ORCID iD
+to your RSA key, thereby making all your publications linkable to you.
+Here is an [example introductory nanopublicaiton](http://purl.org/np/RAy1CYBfBYFd_TFI8Z_jr3taf6fB9u-grqsKyLzTmMvQI).
+
+The link to this nanopublication is also stored in your profile.
+
+## Dependencies
+The ```nanopub``` library currently uses the [```nanopub-java```](https://github.com/Nanopublication/nanopub-java)
```
raar1, line 43: If you look at the ```nanopub/docs/_build/html/setup.html#dependencies``` section of the built docs, the ```nanopub``` and ```nanopub-java``` have exactly the same look/styling, so you can't see that one of them is a link. Is there some way to make that obvious, or perhaps we just add the url explicitly there too?

raar1 (approved):
Very nice, good to have proper docs! I've left some comments/suggestions.

docs/publishing/publishing-nanopublications.md
svenvanderburg: This is a really good point! In the examples I now use `use_test_server=True` but I actually use production URIs so we don't have that confusion there. But this is definitely something users are gonna run into so I've added some separate documentation on this in the 'getting started' part. (I grouped 'What are nanopublications?', 'setup', and 'test server' under 'getting started'.

docs/publishing/retraction.md
svenvanderburg: This is a really good point! I now use production server nanopublications as examples everywhere. (but still having `use_test_server` in the code)

docs/setup.md
svenvanderburg: Fixed!

Issue #59, opened 2020-11-25
Test system undocumented


LourensVeen:
This repository doesn't seem to have developer's documentation. In particular, I had to dive into the GitHub Actions definitions to figure out how to run the test suite. This should be documented in a CONTRIBUTING.rst file.

Can I perhaps suggest that you rebase the repository onto the [NLeSC Python template](https://github.com/NLeSC/python-template)? It's easier if you use it from the start, but I have successfully ported an existing repository over before. It will supply you with a bunch of pre-made configuration, saving a lot of time. And it makes it so that running the test suite is a single `python setup.py test` rather than a whole list of commands, which makes development a lot easier.

raar1:
Good point, yes. Although I note that the NLeSC Python template's build status is failing, which I believe is contrary to best practices ;)

raar1:
@LourensVeen I suspect we may need to update the NLeSC template at some point, if I'm understanding things correctly. I get the following message with ```python setup.py test```:

```
 python3 setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
```

This seems to be related to issues like this:
[https://github.com/pypa/setuptools/issues/1684](https://github.com/pypa/setuptools/issues/1684)

Not sure what to choose instead, really. We could just say 'type ```pytest``` to run tests' instead?

LourensVeen:
Ah, interesting! I have actually run into problems with `pytest-mypy` with this, since `setup.py test` uses easy_install to install the dependencies, and easy_install seems to not understand the `py.typed` file which signals that the module has type annotations, so `pytest-mypy` cannot check the types, which is annoying. So a replacement that uses `pip` would be fine with me.

I think this should be picked up for the template though, if `setup.py test` is deprecated then we should have a new standard for NLeSC, and the template should be updated to match that. And then we can do the same in our repos. So shall we make an issue for this in the template repo?

raar1:
Good idea - I can just add a modified version of what I posted above as an issue perhaps?

raar1:
OK, I've raised an issue here: [https://github.com/NLeSC/python-template/issues/72](https://github.com/NLeSC/python-template/issues/72). Let's see what others say.

LourensVeen:
[...]

PR #67, opened 2020-11-30
Issue59 test system undocumented

raar1:
Addressing issue #59


CONTRIBUTING.md
```diff
@@ -0,0 +1,39 @@
...
+    - some identifying information (name and version number) for dependencies you're using;
+    - information about the operating system;
+1. apply relevant labels to the newly created issue.
+
+## You want to make some kind of change to the code base
+
+1. (**important**) announce your plan to the rest of the community _before you start working_. This announcement should be in the form of a (new) issue;
+1. (**important**) wait until some kind of consensus is reached about your idea being a good idea;
+1. if needed, fork the repository to your own Github profile and create your own feature branch off of the latest master commit. While working on your feature branch, make sure to stay up to date with the master branch by pulling in changes, possibly from the 'upstream' repository (follow the instructions [here](https://help.github.com/articles/configuring-a-remote-for-a-fork/) and [here](https://help.github.com/articles/syncing-a-fork/));
+1. make sure the existing tests still work by running ``pytest``;
```
svenvanderburg, line 33: Maybe mention something about automated testing here? And flake8?

svenvanderburg (approved):
Cool :) One minor comment, but also fine to leave it like this!

CONTRIBUTING.md
raar1: OK, have done so. I don't want to put too much in since it is currently an open issue with the NLeSC style guide and python template. As Lourens had noted, we probably want to find something with the simplicity of ``python setup.py test``` for local testing, instead of pasting in these longer commands.

PR #68, opened 2020-12-01
Add reference guide

svenvanderburg:
* Add API reference to user documentation
* Updates docstrings of client, publication, and namespaces modules.

### DIFF
--- a/examples/searching.ipynb
+++ b/examples/searching.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 2,
    "metadata": {},
    "outputs": [],
    "source": [
@@ -12,7 +12,7 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 2,
+   "execution_count": 3,
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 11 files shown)


examples/searching.ipynb
```diff
@@ -2,7 +2,7 @@
  "cells": [
```
svenvanderburg, line 2: I only added some cells for playing with `search_things`

raar1: Yeah, ipynb files never play nice with git diff...

nanopub/client.py
```diff
@@ -72,13 +89,20 @@ def find_nanopubs_with_pattern(self, subj=None, pred=None, obj=None,
...
+            type (str): A URI denoting the type of the introduced concept
+            searchterm (str): The term that you want to search on
+            max_num_results (int): Maximum number of result, default = 1000
+
         """
-        Searches the nanopub servers (at the specified grlc API) for any nanopubs of the given type, with given search term,
-        up to max_num_results.
-        """
-        if not type or not searchterm:
+        if not searchterm:
```
raar1, line 105: Do we not need to check that type is also specified? If no longer true, then the error message should be clarified.

raar1 (approved):
Cool. Looks like docstrings are generally all up to date now.

nanopub/client.py
svenvanderburg: Good point. I removed the default for `type` argument which makes that check redundant. But indeed the error message is now misleading. I updated it. We should ideally do more validation, but I think this is fine for now.

Issue #62, opened 2020-11-25
CI not enabled for pull requests


LourensVeen:
The `nanopub.yml` workflow only triggers on pushes, not on pull requests. This is no problem for a PR from a local branch, since it will run on push for that, but it doesn't check PRs from a fork of the repo, e.g. PRs made by external contributors.

LourensVeen:
Also, you may want to only push coverage to coveralls for the `main` branch, not for branch commits of PRs.

svenvanderburg:
I disabled it because it would do double builds on pushes to pull requests, but I didn't think of external contributors (even though we are of course expecting loads of them)

Issue #61, opened 2020-11-25
Running pytest --cov overwrites ~/.nanopub


LourensVeen:
Some component of the test suite overwrites the user's profile in ~/.nanopub. This is probably not good, considering that it contains their keys. Also, it makes it impossible to run the test suite twice, unless you recreate your profile, because the test suite needs a valid profile in ~/.nanopub to run.

Running the test suite should not touch anything in the user's home directory, read or write. Use pytest's tmpdir!

svenvanderburg:
Yes, this is really bad...

PR #70, opened 2020-12-01
Fix small stuf

svenvanderburg:
Solves #61, and #62


.github/workflows/build.yml
```diff
@@ -3,9 +3,7 @@
 
 name: Python application
 
-on:
-  push:
-    branches: [ '**' ]
+on: [push, pull_request]
```
svenvanderburg, line 6: This will do a double build on pushes to a pull request, but this cannot be circumvented: https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662

raar1: That's fine, our test suite doesn't take long

raar1 (approved):
No complaints.

PR #71, opened 2020-12-02
Fix flake8 complaints, and make it fail on builds in github action

### LINKED ISSUE #60: Second flake8 command failing silently
The second flake8 command in the CI test suite fails, but this does not mark the result as failed so there are formatting errors in the code that go undetected.

It seems like this may be intentional, but in that case maybe just remove it? What's the point of this check?

svenvanderburg:
Mainly includes a lot of small flake8 complaints, mostly no-brainers, so no need to thoroughly review.
I did some refactoring in publication.py that could have an extra pair of eyes on it though!

Fixes #60


.github/workflows/nanopub.yml
```diff
@@ -26,10 +26,7 @@ jobs:
       - name: Lint with flake8
         run: |
           pip install flake8
-          # stop the build if there are Python syntax errors or undefined names
-          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
-          # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+          flake8 . --count --show-source --statistics
```
svenvanderburg, line 29: I moved configuration to `tox.ini` file. That makes it easier to run it locally using the same settings. (maybe also handy for your CONTRIBUTING related changes)

nanopub/publication.py
```diff
@@ -121,83 +100,119 @@ def from_assertion(cls, assertion_rdf: rdflib.Graph,
                                  'npx:introduces predicate, so you can not also use the '
                                  'introduces_concept argument')
 
-        # To be replaced with the published uri upon publishing
-        this_np = rdflib.Namespace(DUMMY_NANOPUB_URI + '#')
+    @classmethod
+    def from_assertion(cls, assertion_rdf: rdflib.Graph,
```
svenvanderburg, line 104: I completely refactored `from_assertion` method in an attempt to reduce the complexity. I actually really like the result, so I suggest to also let the build fail on complexity violations.

nanopub/publication.py
```diff
@@ -8,6 +7,9 @@
 from nanopub import namespaces, profile
 from nanopub.definitions import DUMMY_NANOPUB_URI
 
+# To be replaced with the published uri upon publishing
+DUMMY_NAMESPACE = rdflib.Namespace(DUMMY_NANOPUB_URI + '#')
```
svenvanderburg, line 11: I realised this is actually constant now. Makes it much cleaner!

tox.ini
```diff
@@ -0,0 +1,5 @@
+[flake8]
+max-complexity = 10
+max-line-length = 100
```
svenvanderburg, line 3: I would vote for this, and make build fail if these are not met to keep things tidy. But I am curious what your opinion is?

.github/workflows/nanopub.yml
raar1: Nice, yes that does. I'll open a new issue to change it.

raar1: Issue #74

nanopub/publication.py
raar1: 👍

nanopub/publication.py
```diff
@@ -51,50 +54,24 @@ def _replace_blank_nodes(dummy_namespace, rdf):
...
-        """
-        Construct Nanopub object based on given assertion. Any blank nodes in the rdf graph are
-        replaced with the nanopub's URI, with the blank node name as a fragment. For example, if
-        the blank node is called 'step', that would result in a URI composed of the nanopub's (base)
-        URI, followed by #step.
+    @staticmethod
+    def _validate_from_assertion_arguments(
+            introduces_concept: rdflib.term.BNode, derived_from, assertion_attributed_to,
+            attribute_assertion_to_profile: bool, provenance_rdf: rdflib.Graph,
+            pubinfo_rdf: rdflib.Graph):
```
raar1, line 67: Should this not also have a docstring? (even though it is indeed an "internal" method)

nanopub/publication.py
raar1: I agree this method is much cleaner now, good job! I have no _strong_ objection to making complexity a fail condition, but in my experience it can sometimes be a bit cumbersome and break the flow. However, clearly we've been ignoring it until now, so let's indeed make it a fail unless/until it becomes too annoying. I think we should keep max line length a non-fail warning though, since that genuinely is not important enough to do every single PR.

nanopub/publication.py
```diff
@@ -121,83 +100,119 @@ def from_assertion(cls, assertion_rdf: rdflib.Graph,
...
-                introduces_concept = this_np[str(introduces_concept)]
-            else:
-                introduces_concept = rdflib.URIRef(introduces_concept)
+        return cls(rdf=main_graph)
 
-            pubinfo.add((this_np[''],
-                          namespaces.NPX.introduces,
-                          introduces_concept))
+    @staticmethod
+    def _handle_assertion_attributed_to(assertion_attributed_to, provenance):
```
raar1, line 193: These utility methods definitely make ```from_assertion()``` easier to parse by eye, so we should keep them. My only issue is that now all these 'private' methods are cluttering the ```Publication``` class. Then again, I think moving them into their own module is more confusing, since the code is more understandable if these helpers are close to the (only) method they act in. Perhaps give them a docstring saying 'handler for from_assertion()' to make it clear which method they belong to.

tox.ini
raar1: Complexity I'm happy with, but ```max-line-length``` can be frustrating for very little gain in readability. Then again, 100 is probably fine (80 is absolutely awful). I think there are cases where a line could be a bit longer and be _clearer_ than when it gets split over several lines, and that's fine as long as it is only occasional. I wish you could have it fail if too many lines are really long...

raar1: OK, thinking about it, let's keep your change as is, and change back if line length gets too bad.

raar1 (approved):
Cleaner code everywhere :) Good stuff. A couple of minor remarks below.

nanopub/publication.py
svenvanderburg: I've added one:         'Validate arguments for `from_assertion` method.' 
I usually don't add them if they are literally the same as the name of a function. But it's good practice anyway.

tox.ini
svenvanderburg: Yeah can always add an `ignore` comment if you have complexity complaints that you don't want to fix.

I am raised with line-length of 80, so 100 already hurts my eyes 😛 🙈 I believe it keeps everything tidy, and if you're used to it it is no effort at all. But indeed let's change it back if it bothers further development!

nanopub/publication.py
svenvanderburg: It depends on how you view it. This code was already in there, only now it is clearer what it does. Indeed it makes the class more cluttered, but the `from_assertion` method less.  But I added the 'handler for from_assertion' docstrings that is a nice suggestion!

Issue #80, opened 2020-12-04
Get test coverage up before release


raar1:
Not a huge priority, but it would be nice for v1.0 to have test coverage of >= 95% or close to that.

raar1:
That was laborious... 92.74% will do for this release!

PR #81, opened 2020-12-04
Issue80 increase test coverage

raar1:
Addressing issue #80


nanopub/java_wrapper.py
```diff
@@ -73,5 +73,12 @@ def _get_signed_file(unsigned_file: str):
...
-        subprocess.run([NANOPUB_SCRIPT, 'mkkeys', '-a', 'RSA'])
+    def make_keys(path_name='~/.nanopub/id'):
+        """
+        Use nanopub-java to make the RSA keys for this user.
+        By default, this uses the path name ~/.nanopub/id and produces a key-pair:
+            ~/.nanopub/id_rsa and ~/.nanopub/id_rsa.pub
+
+        NOTE THAT THE JAVA TOOL ADDS _rsa TO THE END OF YOUR PATH.
+        """
+        subprocess.run([NANOPUB_SCRIPT, 'mkkeys', '-a', 'RSA', '-f', path_name], check=True)
```
svenvanderburg, line 84: I guess this is only necessary for testing?

nanopub/setup_profile.py
```diff
@@ -15,6 +15,7 @@
 
 PRIVATE_KEY_FILE = 'id_rsa'
 PUBLIC_KEY_FILE = 'id_rsa.pub'
+DEFAULT_NP_MKKEYS_PATH = USER_CONFIG_DIR / 'id'
```
svenvanderburg, line 18:
```suggestion
DEFAULT_KEYS_PATH_PREFIX = USER_CONFIG_DIR / 'id'
```

svenvanderburg (approved):
Looks good! I have to say this smells a bit like testing for the sake of testing, but we might thank ourselves for this in the future!

nanopub/java_wrapper.py
svenvanderburg: I was just wondering why this is.

raar1: It does make the testing easier, but it's also to sanitize the tool a little bit. The comments and code all imply that the USERCONFIG directory can be something other than ```~/.nanopub/```, and allow it to be changed. However, by default the nanopub java tool always makes keys in this default directory, which is unexpected behaviour IMO. I've made it use the actual ```USERCONFIG``` dir for everything profile related now.

PR #91, opened 2020-12-11
Catch FileNotFoundError when profile file does not exist

### LINKED ISSUE #87: Catch FileNotFoundError if profile is not setup yet
We do not check for FileNotFoundErrors here: (but they are raised if user did not setup profile)
https://github.com/fair-workflows/nanopub/blob/5a142abae96fc48cd5255cb5781aab6f3f670bc6/nanopub/profile.py#L78

svenvanderburg:
* Fixes #87 
* Also improves error message by pointing to documentation instead of Readme

### DIFF
--- a/nanopub/profile.py
+++ b/nanopub/profile.py
@@ -6,6 +6,11 @@ import yatiml
 
 from nanopub.definitions import PROFILE_PATH
 
+PROFILE_INSTRUCTIONS_MESSAGE = '''
+    Follow these instructions to correctly setup your nanopub profile:
+    https://nanopub.readthedocs.io/en/latest/getting-started/setup.html#setup-your-profile
+'''
+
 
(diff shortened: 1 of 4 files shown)


nanopub/profile.py
```diff
@@ -75,9 +80,9 @@ def get_profile() -> Profile:
     """
     try:
         return _load_profile(PROFILE_PATH)
-    except yatiml.RecognitionError as e:
-        msg = (f'{e}\nYour profile has not been set up yet, or is not set up correctly. To set'
-               f' up your profile, see the instructions in README.md.')
+    except (yatiml.RecognitionError, FileNotFoundError) as e:
```
svenvanderburg, line 83: This is the actual fix for #87

raar1 (approved):
Looks good. Error is clearer now.

Issue #20, opened 2020-10-26
Add methods to NanopubClient for all possible API endpoints


svenvanderburg:
See http://grlc.nanopubs.lod.labs.vu.nl//api/local/local/#/, currently we only support find_text, find_pattern, and find_things. We should add support for all the API endpoints!

raar1:
We no longer provide a one-to-one mapping of the client's methods to the grlc endpoints anyway, so this issue is a bit out-of-date. However, ```NanopubClient._query_grlc()``` now exists, so in principle you can search any available endpoint on the nanopub servers, which essentially solves this issue.

PR #93, opened 2020-12-11
Expand functionality on knowing about retractions

svenvanderburg:
### Added
* `pubkey` option to methods of `NanopubClient` that allows searching for publications 
signed with the given pubkey. For these methods:
    - `find_nanopubs_with_text`
    - `find_nanopubs_with_pattern`
    - `find_things`
* `filter_retracted` option to methods of `NanopubClient` that allows searching for publications 
    that are not retracted. For these methods:
    - `find_nanopubs_with_text`
    - `find_nanopubs_with_pattern`
    - `find_things`
* `NanopubClient.find_retractions_of` method to search retractions of a given nanopublication.
* `Publication.signed_with_public_key` property: the public key that the publication was signed with.
* `Publication.is_test_publication` property: denoting whether this is a publicaion on the test server.

I did not add a `is_retracted` property to the `Publication` class, because we would have to instantiate a `NanopubClient` in the `Publication` object. Then you get all kinds of circular import problems, and the problem that you cannot specify whether you want to use `test` or `prod` server. So users just have to use `NanopubClient.find_retractions_of` to know whether a publication was retracted (I added documentation on this).

### DIFF
--- a/nanopub/client.py
+++ b/nanopub/client.py
@@ -44,7 +44,8 @@ class NanopubClient:
         else:
             self.grlc_urls = NANOPUB_GRLC_URLS
 
-    def find_nanopubs_with_text(self, text: str, pubkey: str = None, max_num_results: int = 1000):
+    def find_nanopubs_with_text(self, text: str, pubkey: str = None,
+                                filter_retracted: bool = True, max_num_results: int = 1000):
         """Text search.
 
         Search the nanopub servers for any nanopubs matching the
@@ -53,6 +54,8 @@ class NanopubClient:
         Args:
             text (str): The text to search on
             pubkey (str): Public key that the matching nanopubs should be signed with
+            filter_retracted (bool): Toggle filtering for publications that are
+                retracted. Default is True, returning only publications that are not retracted.
             max_num_results (int): Maximum number of result, default = 1000
 
         Returns:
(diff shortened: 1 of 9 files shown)


nanopub/client.py
```diff
@@ -62,40 +67,18 @@ def find_nanopubs_with_text(self, text: str, max_num_results: int = 1000):
         """
         if len(text) == 0:
             return []
-
+        endpoint = 'find_signed_nanopubs_with_text'
```
svenvanderburg, line 70: We actually never use the 'normal' find_nanopubs_with_text endpoint anymore.

raar1: That's fine. I don't think unsigned nanopubs are really useful, given the whole point of the infrastructure being to assign authorship/reputation to assertions.

nanopub/client.py
```diff
@@ -180,44 +142,57 @@ def find_things(self, type: str, searchterm: str = ' ',
...
+        else:
+            public_key = None
+
+        results = self.find_nanopubs_with_pattern(pred=namespaces.NPX.retracts,
+                                                  obj=rdflib.URIRef(uri),
+                                                  pubkey=public_key,
+                                                  filter_retracted=False)
+        return [result['np'] for result in results]
 
     def _query_grlc(self, params, endpoint):
```
raar1, line 197: I'd say this essentially solves #20 (as much as it will ever be solved)

nanopub/publication.py
```diff
@@ -28,10 +29,11 @@ class Publication:
         source_uri (str): The URI of the nanopublication that this Publication represents (if
             applicable)
         introduces_concept: The concept that is introduced by this Publication.
-
+        signed_with_public_key: The public key that this Publication is signed with.
+        is_test_publication: Whether this is a test publication
     """
 
-    def __init__(self, rdf=None, source_uri=None):
+    def __init__(self, rdf, source_uri=None):
```
raar1, line 36: Should we give a warning/error if the ```rdf``` passed is e.g. ```None```?

raar1 (approved):
Great PR. Documentation is good, and extensive tests. I've left a couple of comments/questions.

nanopub/publication.py
svenvanderburg: This now gives an error if you don't specify it, which is already a huge step forward. 

Checking that the `rdf` is not `None` would be nice, we could also check that it is an instance of `rdflib.Graph`. We don't really do much type & value checking anywhere throughout the library, but we could. Maybe you should create an issue if you want to do this (but I'm fine with not doing it)

I've added type hinting so it is clear what to pass which I think is enough in this case. We don't expect the user to use this actually.

PR #96, opened 2020-12-15
Add .zenodo.json

svenvanderburg:


raar1 (approved):
Looks fine, but I'm wondering whether the CITATION.cff file could also be referenced in the JSON

svenvanderburg:
You mean to have all information in 1 file instead of 2? I think you need both, and they have different notation so I am not sure if we can include one in the other. The` zenodo.json file` was completely automatically created used `cffconvert` though. People do automate that using github actions, see [this](https://github.com/citation-file-format/cffconvert-github-action). But I think that's a bit too farfetched.

raar1:
Oh, it just felt a bit like duplication of the same info. But I didn't realise it was created automatically, so indeed we can have a github action for it later if we really cared.

Issue #95, opened 2020-12-14
Fix publish coverage error


svenvanderburg:
We had this error:
```
coveralls.exception.CoverallsException: Running on Github Actions but GITHUB_TOKEN is not set. Add "env: GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}" to your step config.
```
on #94

vemonet:
Not sure how your test coverage work, but the error message seems not to be pointing to the right issue

Usually after doing the tests it will should try to publish the test coverage to the website you use (coveralls here apparently)

The action workflow:

```yaml
env:
          COVERALLS_REPO_TOKEN: ${{ secrets.COVERALLS_REPO_TOKEN }}
```

I did not set the secret `COVERALLS_REPO_TOKEN` (and should not, it is for your repo only)

What you can do to make it cleaner probably is to add a `if` for the coverall publishing step, so that the step is run only if the `secrets.COVERALLS_REPO_TOKEN` is set

svenvanderburg:
Fixed by #98

svenvanderburg:
Thanks a lot @vemonet 👍

PR #98, opened 2020-12-15
Add installation guidelines to contributing.md and fix coverall publishing issue in workflow

vemonet:
* Add install step to contributing to make it clear how to install dependencies for development
* Add `if: ${{ env.COVERALLS_REPO_TOKEN }}` to `build.yml` workflow to fix issue when the coveralls token is not defined (for forks for example). Related to issue https://github.com/fair-workflows/nanopub/issues/95

Example of successful run: https://github.com/vemonet/nanopub/actions/runs/423181615

It needs to be tested on the main repo to make sure it does not prevent to properly run the publishing step when the coverall token is defined

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -36,9 +36,9 @@ jobs:
         run: |
           pytest --cov
       - name: Publish coverage to Coveralls
-        if: ${{ secrets.COVERALLS_REPO_TOKEN }}
         env:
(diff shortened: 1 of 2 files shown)


svenvanderburg (approved):
Nice changes! 

I tested that it doesn't skip publishing when the COVERALLS_TOKEN is there in #99.

PR #102, opened 2020-12-17
 nanopub-java installed upon installation instead of runtime

### LINKED ISSUE #88: permission denied when creating nanopub-java jar file
When trying to publish using the nanopub library when it is installed in a non-user directory we get a `RunTimeError` with:
```
Warning: Failed to create the file  
Warning: /usr/local/lib/python3.8/dist-packages/nanopub/nanopub-1.32-jar-with-d ependencies.jar: Permission denied
```
We need to download the `nanopub-java` file to a different location than the installation folder so that we have access to it. A good candidate would be the `~/.nanopub` folder.

svenvanderburg:
This fixes #88, I tested that using the following dockerfile:

```docker
FROM ubuntu:latest

RUN apt-get update
RUN apt-get install -y python3-pip curl
ARG DEBIAN_FRONTEND=noninteractive
RUN apt-get install -y default-jre
RUN apt-get install -y git

RUN pip3 install git+https://github.com/fair-workflows/nanopub.git@clean-java-dependency
RUN useradd -ms /bin/bash sven
```

Then testing manually in interactive terminal:
```
docker build -t test .
docker run -it --user sven test
```
This works fine, whereas using `main` branch results in permission denied error because nanopub is installed as root and we try to download the `.jar` file on runtime.


raar1 (approved):
Very good work. I wonder if we could have that dockerfile you used to test this in the repo as well? Perhaps as another test?

svenvanderburg:
As discussed, it would ideally be nice to use that dockerfile as a test, but for now we can live with this tiny technical debt.

PR #105, opened 2020-12-21
Add pagination of search results

svenvanderburg:


CHANGELOG.md
```diff
@@ -9,9 +9,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 * Added Zenodo badge to README
+* Pagination of results for search methods of `NanopubClient`
 
 ### Changed
 * `nanopub-java` dependency is installed upon installation instead of upon runtime.
+* search methods of `NanopubClient` return iterator instead of list
```
svenvanderburg, line 16: In theory this is a backwards incompatible API change, so we should release to v2.0.0. I also don't want to make this backwards compatible. What do you think @raar1?

tests/test_client.py
```diff
@@ -157,20 +159,6 @@ def test_find_retractions_of_valid_only(self):
         unexpected_uri = 'http://purl.org/np/RACdYpR-6DZnT6JkEr1ItoYYXMAILjOhDqDZsMVO8EBZI'
         assert unexpected_uri not in results
 
-    def test_nanopub_search(self):
```
svenvanderburg, line 160: I removed this, because I think it is redundant.

CHANGELOG.md
raar1: I agree, let's have it as a backwars incompatible change this time round. I think we only have a couple of users so far so it should be ok to make this change and communicate it to them. In future of course (especially after the blog post) we will want to avoid such changes, I guess. I am tempted simply to increment by the usual 0.1, unless you strongly object?

docs/index.rst
```diff
@@ -115,7 +115,7 @@ Searching for nanopublications
 
     # Search for all nanopublications containing the text 'fair'
     results = client.find_nanopubs_with_text('fair')
-    print(results)
+    print(list(results))
```
raar1, line 118: Should we note here that only the first 10000 results will appear in the resultant list? (assuming that is the pagination limit?)

tests/test_client.py
raar1: Agreed.

raar1 (approved):
This is very nice - I like the pagination-via-generator approach, and the simplification of the search method signatures.

docs/index.rst
svenvanderburg: The pagination limit is 1000, but the list will be as large as the number of matching nanopublications (it just eats up the entire generator). But I will change this to:
```python
for result in results:
    print(result)
```

CHANGELOG.md
svenvanderburg: Yes, the new version shall be 1.2.0

PR #107, opened 2020-12-22
Add bin/nanopub-java.bat and include it in installation for windows

svenvanderburg:
Makes package work on Windows
* Added nanopub-java.bat file to run the `nanopub-java` on windows
* Several small fixes in python code to make compatible with windows
* Tested extensively locally on a windows machine, all tests pass, but I could not get it to work in a github action. I created #118 to fix it later.


.github/workflows/build.yml
```diff
@@ -7,31 +7,39 @@ on: [push, pull_request]
 
 jobs:
   build:
-
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [windows-latest]
```
svenvanderburg, line 13: TODO: This should also include ubuntu-latest

nanopub/setup_nanopub_profile.py
```diff
@@ -39,12 +39,11 @@ def validate_orcid_id(ctx, orcid_id: str):
                     '(by default HOMEDIR/.nanopub/). '
                     'The profile will also be published to the nanopub servers.')
 @click.option('--keypair', nargs=2, type=Path,
-              prompt=f'If the public and private key you would like to use are not '
-                     f'in {USER_CONFIG_DIR}, provide them here. '
-                     f'If they are in this directory or you wish to generate new keys, '
-                     f'leave empty.',
               help='Your RSA public and private keys with which your nanopubs will be signed',
               default=None)
+@click.option('--newkeys', type=bool, is_flag=True, default=False,
```
svenvanderburg, line 44: I added this so we can just call `setup_nanopub_profile` fully non-interactively from the command line (which is much nicer for github action).

bin/nanopub-java.bat
```diff
@@ -0,0 +1,76 @@
...
+@rem Get command-line arguments, handling Windows variants
+
+if not "%OS%" == "Windows_NT" goto win9xME_args
+
+:win9xME_args
+@rem Slurp the command line arguments.
+set CMD_LINE_ARGS=
+set _SKIP=2
+
+:win9xME_args_slurp
```
raar1, line 55: This is so... beautiful...

nanopub/setup_nanopub_profile.py
raar1: Yes, this is a much needed option.

raar1 (approved):
Well done for persevering with this nightmare. As discussed before, getting the tests to pass on windows CI is proving very difficult, but it seems to work locally and windows support is not a priority unless/until we have users on that OS, so let's leave it for now and return once we have more time.

raar1:
We should disable the windows CI tests for now, before merging.

PR #113, opened 2020-12-28
Rename tox.ini to setup.cfg to avoid confusion

svenvanderburg:


LourensVeen:
Good idea! :-)

LourensVeen (approved):
Good idea!

Issue #111, opened 2020-12-24
JSONDecodeError when search result is empty


svenvanderburg:
```python
results = client.find_nanopubs_with_text('some random string that is not in any of the nanopublications')
print(next(results))
```
```python
---------------------------------------------------------------------------
JSONDecodeError                           Traceback (most recent call last)
<ipython-input-168-bab0d5cfb459> in <module>
      1 results = client.find_nanopubs_with_text('some random string that is not in any of the nanopublications')
----> 2 print(next(results))

~/projects/fair-workflows/nanopub/nanopub/client.py in _search(self, endpoint, params)
    246                 r = self._query_grlc(params, endpoint, grlc_url)
    247                 r.raise_for_status()
--> 248             results = r.json()
    249             bindings = results['results']['bindings']
    250             if not bindings:

~/opt/anaconda3/envs/fair3.6/lib/python3.6/site-packages/requests/models.py in json(self, **kwargs)
    896                     # used.
    897                     pass
--> 898         return complexjson.loads(self.text, **kwargs)
    899 
    900     @property

~/opt/anaconda3/envs/fair3.6/lib/python3.6/json/__init__.py in loads(s, encoding, cls, object_hook, parse_float, parse_int, parse_constant, object_pairs_hook, **kw)
    352             parse_int is None and parse_float is None and
    353             parse_constant is None and object_pairs_hook is None and not kw):
--> 354         return _default_decoder.decode(s)
    355     if cls is None:
    356         cls = JSONDecoder

~/opt/anaconda3/envs/fair3.6/lib/python3.6/json/decoder.py in decode(self, s, _w)
    337 
    338         """
--> 339         obj, end = self.raw_decode(s, idx=_w(s, 0).end())
    340         end = _w(s, end).end()
    341         if end != len(s):

~/opt/anaconda3/envs/fair3.6/lib/python3.6/json/decoder.py in raw_decode(self, s, idx)
    355             obj, end = self.scan_once(s, idx)
    356         except StopIteration as err:
--> 357             raise JSONDecodeError("Expecting value", s, err.value) from None
    358         return obj, end

JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

raar1:
Yeah, I think the problem here is that it is receving HTML instead of JSON, since virtuoso has given an error (it doesn't like the spaces in the free form text). I've tried to handle this a little more gracefully in PR #117

raar1:
Fixed in PR #117

PR #117, opened 2021-01-08
Issue111 json decode error empty search result

raar1:
Address #111

### DIFF
--- a/nanopub/client.py
+++ b/nanopub/client.py
@@ -252,7 +252,11 @@ class NanopubClient:
             try:
                 results = r.json()
             except ValueError:
-                return
+                # Try to give a more understandable error to user when the response
+                # is not JSON...
+                raise ValueError('The server returned HTML instead of the requested JSON.'
+                                 'This is usually caused by the triple store (e.g. virtuoso)'
+                                 'throwing an error for the given search query.')
 
             bindings = results['results']['bindings']
             if not bindings:
--- a/nanopub/client.py
+++ b/nanopub/client.py
@@ -245,7 +245,15 @@ class NanopubClient:
             else:
                 r = self._query_grlc(params, endpoint, grlc_url)
                 r.raise_for_status()
-            results = r.json()
+
+            # Check if JSON was actually returned. HTML can be returned instead
+            # if e.g. virtuoso errors on the backend (due to spaces in the search
+            # string, for example).
+            try:
+                results = r.json()
+            except ValueError:
+                return
+
             bindings = results['results']['bindings']
             if not bindings:
                 has_results = False
--- a/nanopub/client.py
+++ b/nanopub/client.py
@@ -251,12 +251,12 @@ class NanopubClient:
             # string, for example).
             try:
                 results = r.json()
-            except ValueError:
+            except ValueError as e:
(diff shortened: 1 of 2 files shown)


nanopub/client.py
```diff
@@ -245,7 +245,15 @@ def _search(self, endpoint: str, params: dict):
...
                 r = self._query_grlc(params, endpoint, grlc_url)
                 r.raise_for_status()
-            results = r.json()
+
+            # Check if JSON was actually returned. HTML can be returned instead
+            # if e.g. virtuoso errors on the backend (due to spaces in the search
+            # string, for example).
+            try:
+                results = r.json()
+            except ValueError:
```
svenvanderburg, line 254: Why ValueError and not JSONDecodeError?

svenvanderburg (approved):
Nice addition!

nanopub/client.py
raar1: Mainly because the online examples used it, possibly to avoid needing to consider what json lib ```requests``` is using. But since it is obviously going to use ```json``` from now on then we can use JSONDecodeError if you prefer?

I think I still like the idea of catching similar exceptions created during the conversion to JSON since a ```nanopub``` user will not understand or be able to do anything about that. Maybe we should dump it all to an error log that users can access if they want insight into why server queries are failing?

LourensVeen: This API seems a bit funny to me. Is it designed for human interaction, or for machine interaction?

If the former, then it should communicate in HTML, if the latter, then in JSON. But now it seems that it communicates data in machine-readable format if things are well (which is good if there's a script on the other side, and useless if there's a human), and in human-readable format if things are not well (which is good if there's a human on the other side, but useless to a script).

So it seems that the idea is that a script queries it, and if there's an error, fires up a browser to show a user what's wrong? Assuming that a browser makes sense in whatever environment the script was started in? What Python library does that?

The standard way of signalling an error in Python is by raising an exception. So that's what the nanopub library should do. And then I guess it just needs to do its best to give as good an error message as it can, given the limitations imposed by the server. If the HTML is simple, maybe you can strip the tags and put whatever remains in the exception message? Or parse it and extract a specific element?

svenvanderburg: @LourensVeen there is nothing we can do about the nanopub server returning us HTML. But we currently return `None` to the user in that case which is strange. We should give a more informative error message.

@raar1 cool, let's keep using `ValueError`!

raar1: I agree that it is ugly, but as Sven says we don't really have much control over this behaviour. I would like to convert the exception to a warning, effectively. Something along the lines of ```Warning: Server did not return JSON as expected. This query caused an error.```

Perhaps what we should do is catch the 'bad' text string before we send it? Ideally though, the server would always return JSON if asked for it, so we could instead see if the people running the different nanopub servers could agree on a standard response? I think in principle they might not all be using virtuoso as the triple store, so this 'free form text' error might not always happen...

svenvanderburg: I quite like the message in that warning, but I would make it throw an actual error... I don't think we should do more than that.

raar1: Done.

Issue #119, opened 2021-01-21
Some valid ORCIDs do not match the regex check


raar1:
From a prospective user of the library:

```
hi, i got stuck setting up my nanopub profile because my ORCID does not
match the regex. I think the regex should be : ORCID_ID_REGEX = r'^https://orcid.org/(\d|X{4}-){3}\d|X{4}$' to allow for the Xes"

ORCIDs can end with an "X" instead of a number.

This is somehow because the last digit is a checksum.
```

raar1:
Fixed by PR #121

PR #121, opened 2021-01-21
Issue119 orcid regex

raar1:
Addressing #119 

Regex should now allow orcids ending in X. This does not do the checksum though.


svenvanderburg (approved):
Looks good!

Issue #114, opened 2021-01-05
Potential problem with Nanopub in use with Nanotate


raar1:
See [https://github.com/nanotate-tool/nanotate/issues/1](https://github.com/nanotate-tool/nanotate/issues/1)

Essentially the nanopub is being signed but rejected by the servers with:
```
HTTP Status 400 – Bad Request' with message ''Nanopub is not trusty:"
```

raar1:
So after more digging, Miguel found out the issue had something to do with UTF-8:
[https://github.com/nanotate-tool/nanotate/issues/1#issuecomment-759130568](https://github.com/nanotate-tool/nanotate/issues/1#issuecomment-759130568)

raar1:
The ``` ° ``` character in the text was causing the problem. 

Modifying the run command to
```
exec java -Dsun.jnu.encoding=utf8 -Duser.language=en -Duser.country=US -Dfile.encoding=utf8 -jar $NANOPUBJAR "$@"
```
seemed to fix the issue for him. Output of his ```java --version``` is:
```
java version "1.8.0_211"
Java(TM) SE Runtime Environment (build 1.8.0_211-b12)
Java HotSpot(TM) 64-Bit Server VM (build 25.211-b12, mixed mode)
```

raar1:
No idea if this is a portable solution, but we could look into adding some of these flags to cover more installation possibilities? What do you think @svenvanderburg ?

svenvanderburg:
Oh my god, this is horrible... Yeah we could just hardcode these flags in `bin/nanopub-java` and `bin/nanopub-java.bat` script. Or de we expect users to configure this themselves? (I wouldn't)

raar1:
I really wouldn't expect users to configure this themselves, no. I'm just worried these flags might break for other versions of java, or something like that. We can try with these flags and check at least that it still works for us too.

raar1:
This was fixed by PR #122

PR #122, opened 2021-01-21
Add utf8 flags to java call in nanopub-java

raar1:
Addressing #114 

Adds utf8 flags to ```nanopub-java``` bash script.


svenvanderburg (approved):
Let's hope that this error only occurs on unix...

PR #125, opened 2021-01-25
Add azure devops sync github action

svenvanderburg:


svenvanderburg:
I can only test this in main, so I will merge.

PR #141, opened 2021-08-09
Add license for nanopub-java

svenvanderburg:
I think this is needed to add nanopub to conda-forge, see https://github.com/conda-forge/staged-recipes/pull/15870.


svenvanderburg:
Now that I think of it, the license is ofcourse included in the .jar file.

Issue #110, opened 2020-12-24
Fetching a publication, modifying, and republishing fails


svenvanderburg:
## How to reproduce
```python
>>> publication = client.fetch('http://purl.org/np/RApJG4fwj0szOMBMiYGmYvd5MCtRle6VbwkMJUb1SxxDM')
>>> publication.assertion.add((rdflib.URIRef('http://example.org/timbernerslee'),
>>>                                               rdflib.RDF.type,
>>>                                               rdflib.FOAF.Person))     
>>> client.publish(publication)
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-64-86cb32f8e4cf> in <module>
----> 1 client.publish(publication)

~/projects/fair-workflows/nanopub/nanopub/client.py in publish(self, publication)
    318 
    319         # Sign the nanopub and publish it
--> 320         signed_file = self.java_wrapper.sign(unsigned_fname)
    321         nanopub_uri = self.java_wrapper.publish(signed_file)
    322         publication_info = {'nanopub_uri': nanopub_uri}

~/projects/fair-workflows/nanopub/nanopub/java_wrapper.py in sign(self, unsigned_file)
     43     def sign(self, unsigned_file: Union[str, Path]) -> str:
     44         unsigned_file = str(unsigned_file)
---> 45         self._run_command(f'{NANOPUB_JAVA_SCRIPT} sign ' + unsigned_file)
     46         return self._get_signed_file(unsigned_file)
     47 

~/projects/fair-workflows/nanopub/nanopub/java_wrapper.py in _run_command(command)
     39                                + PROFILE_INSTRUCTIONS_MESSAGE)
     40         elif result.returncode != 0:
---> 41             raise RuntimeError(f'Error in nanopub java application: {stderr}')
     42 
     43     def sign(self, unsigned_file: Union[str, Path]) -> str:

RuntimeError: Error in nanopub java application: java.security.SignatureException: Seems to have trusty URI before signing: http://purl.org/np/RApJG4fwj0szOMBMiYGmYvd5MCtRle6VbwkMJUb1SxxDM
	at org.nanopub.extra.security.SignNanopub.signAndTransform(SignNanopub.java:150)
	at org.nanopub.extra.security.SignNanopub.signAndTransform(SignNanopub.java:144)
	at org.nanopub.extra.security.SignNanopub.writeAsSignedTrustyNanopub(SignNanopub.java:203)
[...]

PR #142, opened 2021-08-10
Add userfriendly error message in case of signing an already signed file

svenvanderburg:
Related to #110

### DIFF
--- a/nanopub/java_wrapper.py
+++ b/nanopub/java_wrapper.py
@@ -39,7 +39,7 @@ class JavaWrapper:
                                + PROFILE_INSTRUCTIONS_MESSAGE
                                + '\nDetailed error message:\n' + stderr)
         elif all(m in stderr for m in ['SignatureException', 'Seems to have signature']):
-            raise RuntimeError('The Publication you try to publish already has a signature, '
+            raise RuntimeError('The Publication you are trying to publish already has a signature, '
                                'this means it is likely already published. '
                                'If you want to publish a modified existing nanopublication '
                                'you need to do a few extra steps before you can publish. '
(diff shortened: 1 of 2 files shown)


nanopub/java_wrapper.py
```diff
@@ -38,6 +38,13 @@ def _run_command(command):
             raise RuntimeError('Nanopub RSA key appears to be missing,\n'
                                + PROFILE_INSTRUCTIONS_MESSAGE
                                + '\nDetailed error message:\n' + stderr)
+        elif all(m in stderr for m in ['SignatureException', 'Seems to have signature']):
+            raise RuntimeError('The Publication you try to publish already has a signature, '
```
raar1, line 42:
```suggestion
            raise RuntimeError('The Publication you are trying to publish already has a signature, '
```

tests/test_java_wrapper.py
```diff
@@ -40,6 +40,15 @@ def test_sign_fails_on_invalid_nanopub(tmp_path):
         java_wrapper.sign(invalid_file)
 
 
+def test_sign_fails_on_already_signed_publication(tmp_path):
+    java_wrapper = JavaWrapper(use_test_server=True)
+    temp_signed_file = tmp_path / 'signed.trig'
+    shutil.copy(NANOPUB_SAMPLE_SIGNED, temp_signed_file)
+    with pytest.raises(RuntimeError) as e:
+        java_wrapper.sign(unsigned_file=temp_signed_file)
+    assert 'The Publication you try to publish already has a signature' in str(e.value)
```
raar1, line 49:
```suggestion
    assert 'The Publication you are trying to publish already has a signature' in str(e.value)
```

raar1 (approved):
A nice, clear error message instead of the java stacktrace of before. I like that it links to the github issue. If we get a user actually trying to do this then we can use their case to help design the solution.

Issue #149, opened 2022-09-15
Making the addition of `prov:generatedAtTime` and `prov:wasAttributedTo` optional


vemonet:
The `nanopub` python library is automatically adding `prov:generatedAtTime` duplicated twice in the pubinfo and prov graph. And a `prov:wasAttributedTo` in the pubinfo. 

I added 2 parameters to the `Publication.from_assertion` method: `add_generated_at_time` (default to True) and `attribute_publication_to_profile` (default to True): https://github.com/vemonet/nanopub/commit/0df2db6203ab3679334449307cb797535d24693a 

So that the default behavior is still the same (adding those metadata), but users can also choose to add them by themselves

Is that something that you would be interested to see merged here? @svenvanderburg

raar1:
That would be great @vemonet ! I agree that the duplication is not ideal and users should definitely be able to specify these times themselves. Would you be willing to open a PR with your suggested changes and we can review it? Thanks again for your help.

raar1:
@vemonet Thanks again for this. I've had a look at the PR and it looks good. I completely agree that it should be possible to disable the time and profile attributions. What I'm less sure about is what the defaults should be.

Currently in the code if `attribute_publication_to_profile` is False, then `publication_attributed_to` will be ignored. This makes sense, of course, if the user has explicitly said they don't want to attribute the publication to a profile. However, if `attribute_publication_to_profile` is set False by default (which I think you were suggesting could be a better choice) then this might confuse users who do e.g.:
```python
publication = Publication.from_assertion(assertion_rdf=my_assertion, publication_attributed_to=someorcid)
```
and then wonder why `publication_attributed_to` isn't working.

I think either we keep your code as it is, but leave `attribute_publication_to_profile=True` as default. Or we make the use of `publication_attributed_to` override `attribute_publication_to_profile`, and always attribute if it is set.

What do you think?

vemonet:
Hi @raar1, I agree we should leave those defaults. After second thought, having the ORCID of the publisher and the time the nanopub was published added to the pubinfo graph by default is a good practice. The only thing is that it can be a bit surprising when you are used to publish with the java tool that does not add anything

There is also the fact that `prov:generatedAtTime` is automatically added to the pubinfo and to the provenance graphs that feels a bit weird. It would make more sense if it would be only added to the pubinfo. 
But publishing a nanopub with nothing in the provenance graph will fail, so it can be useful to have at least the generatedAtTime in there by default...

[...]

PR #151, opened 2022-09-15
Add options to let users choose if they want `prov:generatedAtTime` and `prov:attributedTo` automatically added to prov/pubinfo graphs

vemonet:
The issue addressed by this pull request has been mentioned in https://github.com/fair-workflows/nanopub/issues/149

## Current state (before this pull request)

When publishing a nanopub the lib was automatically generating and adding:
* `prov:generatedAtTime` duplicated in the pubinfo and prov graph
*  `prov:wasAttributedTo` the user ORCID in the pubinfo

## Changements

I added 2 parameters to `Publication.from_assertion()` to let users choose:
* `add_generated_at_time` (default to True)
* `attribute_publication_to_profile` (default to True)

Currently the 2 are defaulting to `True` to keep the default behavior the same as before.

It does not break any existing test (according to my local testing)

## Question

Should we keep them with True as default? 

To me it would make more sense they are disabled by default, and the user need to explicitly ask for it, like for the others existing parameters (`attribute_assertion_to_profile`, `publication_attributed_to`...).

### DIFF
--- a/nanopub/publication.py
+++ b/nanopub/publication.py
@@ -3,13 +3,13 @@
 This module holds code for representing the RDF of nanopublications, as well as helper functions to
 make handling RDF easier.
 """
-from copy import deepcopy
 import warnings
+from copy import deepcopy
 from datetime import datetime
 from urllib.parse import urldefrag
 
 import rdflib
-from rdflib.namespace import RDF, DC, DCTERMS, XSD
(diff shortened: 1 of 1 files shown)


raar1 (approved):
This is good, I'm going to merge it now. Thanks a lot for contributing this @vemonet !

PR #158, opened 2022-12-15
Fix how introduces_concept is handled

vemonet:
@raar1 While upgrading the `fairworkflow` library I noticed an issue with how `introduces_concept` is handled, especially when generating the concept URI. So here is a fix!


raar1 (approved):
Thanks! Good work - I will merge this last PR before making the 2.0.0 release

PR #162, opened 2023-11-08
Fix searching for nanopubs example in docs

Ostrzyciel:
The example omitted initializing the `client` variable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,6 +68,7 @@ print(np)
 from nanopub import NanopubClient
 
 # Search for all nanopublications containing the text 'fair'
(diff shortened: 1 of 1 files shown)


Ostrzyciel:
I've just noticed that the following example (fetching nanopubs) does not work as well – there is no `fetch` method in `NanopubClient`.

vemonet (approved):

raar1:
Thanks very much @Ostrzyciel ! Apologies for the late replies here - we'll try to get around to fixing a few of the issues over the coming weeks.

PR #165, opened 2024-01-31
Fix nanopub-java version to 1.47

vemonet:
1.47 is the latest version using purl.org (tests fails when using the new w3id.org version because it generates different hashes) @raar1


raar1 (approved):
Thanks @vemonet ! I will merge this as it fixes all the build tests. I think the docs failing happens when it is not in the main branch (can't remember) but I will check and fix that if necessary afterwards

PR #180, opened 2025-05-03
Implement `np create` command

anatoly-scherbakov:
# Why?

At this point, `np sign` or `np publish` require a full fledged nanopublication with correctly specified head, pubinfo, provenance, and assertion graphs. This might be rather perplexing for a user, especially for one who is not a specialist in LD per se.

Meanwhile, `nanopub-py` has a mechanism of templates, which is available through code.

# What

What if to make it available through the CLI as well?

This PR introduces a new command, `np create`, which has a number of subcommands.

```
$ np create

 Usage: np create [OPTIONS] COMMAND [ARGS]...

 Create a Nanopublication from a template and print it to standard output.


╭─ Options ────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ --output-format        [trig|xml|nquads|json-ld]  [default: trig]                                                                                                                                                                    │
│ --help                                            Show this message and exit.                                                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
╭─ Commands ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╮
│ from-assertion   Create a nanopub from an assertion graph                                                                                                                                                                            │
│ claim            Create a nanopublication based on a free-form textual claim.                                                                                                                                                        │
╰──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

These commands will accept arguments, variable for each case, and output a formulated nanopublication to stdout. Thereafter it can be viewed to confirm that it is really what the user wants, or saved to a file using shell redirects.

Whereupon it can be signed and published.

## Examples

```
$ np create claim Dinosaurs like apples.
$ np create --output-format nquads from-assertion stuff.ttl --was-derived-from https://example.org/nonsense
```

## Further development

- Refactor `np create` into a separate file to keep `__main__.py` smaller
- Make templates mechanism extensible so that people could publish Python packages with them. For instance,

```
pip install nanopub-discworld
```

would make new commands available, like `np create spell`, which would describe very domain specific nanopublications. This can be done using setuptools endpoints mechanism.

# Questions

What do you think? I do not know how this aligns with general course of development for the tool. If you think it's ok I would maybe look into adding these changes to the docs. Let me know! Thanks.


VirginiaBalseiro (approved):
:rocket:

anatoly-scherbakov:
Thanks for approving @VirginiaBalseiro!

I don't think I have the rights to merge this PR however. Could you please advise on the next steps?

anatoly-scherbakov:
Thanks @cmeesters!

I am also not a project member, but I have glanced over your PR and it looks good. I left one question there though.

I agree that fixing the CI workflow is an important thing to propel the project forward.

> And still think this deserves a special option, as the script opens up a curses UI

I believe a plugin mechanism would support this. Your plugin could expose, say,

```
nanopub create species --interactively
```

…which could open a textual (or even graphical) UI to compose the assertion interactively.

I have had interesting results writing nanopublications with the help of LLMs, and I believe they can improve productivity of doing so greatly.

Possibilities are endless.

anatoly-scherbakov:
@tkuhn @VirginiaBalseiro hi! Let me know if this branch is mergeable. I can see that coverage CI checks are red. Let me know if this can be addressed in follow-up PRs.

ashleycaselli:
> I can see that coverage CI checks are red. Let me know if this can be addressed in follow-up PRs.

It's on our side because we left the coverage decrease threshold to the default value (0%), which means any decrease in coverage would result in CI failing. I've just increased that to 1% which is a more reasonable value.

PR #193, opened 2025-06-10
fix: introducing return values for publish()

cmeesters:
motivation: 

Whilst the publish() method logs URI and server for a terminal user to read, programmatically there is no way to get hold of these values. Hence, they ought to be returned, too. Otherwise, a user cannot check on return values and in scripts logging output might be supressed.


VirginiaBalseiro (approved):
looks good!

PR #207, opened 2025-08-09
fix: ci

cmeesters:
This PR is merely an attempt to 
- set dependencies up to date and
- to fix the CI issues

### DIFF
--- a/.github/workflows/pypi.yml
+++ b/.github/workflows/pypi.yml
@@ -8,12 +8,12 @@ jobs:
   publish:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+    - uses: actions/checkout@v4
 
     - name: Set up Python
-      uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
+      uses: actions/setup-python@v5
       with:
-        python-version: "3.7"
+        python-version: "3.11"
 
     - name: Install build dependencies
       run: pip install build
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,9 +4,10 @@ build-backend = "hatchling.build"
 
 [project]
 name = "nanopub"
+version = "2.0.1"
 description = "Python client for Nanopublications"
 readme = "README.md"
-requires-python = ">=3.12.0"
+requires-python = ">=3.12"
 license = { file = "LICENSE" }
 authors = [
     { name = "Robin Richardson", email = "r.richardson@esciencecenter.nl" },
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -70,11 +72,11 @@ np = "nanopub.__main__:cli"
...
-Homepage = "https://nanopublication.github.io/nanopub-py/"
-Documentation = "https://nanopublication.github.io/nanopub-py/"
-History = "https://github.com/Nanopublication/nanopub-py/releases"
-Tracker = "https://github.com/Nanopublication/nanopub-py/issues"
-Source = "https://github.com/Nanopublication/nanopub-py"
+Homepage = "https://fair-workflows.github.io/nanopub"
+Documentation = "https://fair-workflows.github.io/nanopub"
+History = "https://github.com/fair-workflows/nanopub/releases"
+Tracker = "https://github.com/fair-workflows/nanopub/issues"
+Source = "https://github.com/fair-workflows/nanopub"
```
anatoly-scherbakov, line 79: These links redirect to `https://github.com/Nanopublication/nanopub-py/…` as far as I can see; I am not sure what `fair-workflows` is. Maybe more context would be needed?

anatoly-scherbakov (approved):
Changes look fair to me (although I did not live test this PR), there's one question though.

pyproject.toml
cmeesters: I made a comment, that this is to be considered a draft PR. I need it to be labelled a conventional PR, for otherwise the pipelines will not run. The toml file is huge and contains numerous issues. I will work on it as time permits.

ashleycaselli (changes requested):
Besides the comments I left on the single lines, in my opinion, this PR includes too many changes that are out of the scope of the PR. Indeed, it is named `fix:ci` but then
- it introduces `poetry` as build system
- makes a major code refactor

This might be misleading and also I don't think it's a good idea to mix up things such as fixing the broken CIs and changing the project build system. We can discuss how to bring this further but I do think we should address the issues with different PRs.

.github/workflows/pypi.yml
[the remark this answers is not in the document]
cmeesters: Hi, 

Thanks for your comment!

> makes a major code refactor

I hardly changed anything beyond the CI parts and toml file - the apparent changes are resulting from a run of `black` to adjust formatting. This is, indeed, a matter of taste.

The CI, has indeed been substantially refactored:
- due to issues installing `isort` into this environment, I changed to linting / formatting supported by poetry.
- I removed some deprecated functions and inserted alternate marks, keeping the functionality.
- made a difference between flaky and non-flaky tests in the CI. In my view, downstream work would need to address some issues

That is it. Basically. All tests can be run locally, before submitting using poetry. Again, this is a matter of taste. Yet, the functionality has been preserved. Along the way, I discovered some more deprecated stuff (see #208). Also, the formatting still seems broken. It can be tuned and be a little less picky. If you like me to, I can easily adjust this. Unused variables are a bit tricky, though. Better to correct for such issues.

Regarding the URLs and the action versioning: I can do this, too. However, I plan to write a Snakemake plugin based on this library and hence, need a reliable CI and updating scheme. This is paramount for further contributions, like the one mentioned in #180. While I do not share your view on the GitHub action security, this, too, can be reverted / updated. To continue, I would like you or other to review and comment, whether there is interest in this PR (even if only for cherry-picking).

anatoly-scherbakov:
I personally do not like `black` too much but I do not have much against it either. I believe auto formatting code does not need manual review. This might mean that a separate PR could be used to house all changes ensuing from running a formatter against the codebase.

ashleycaselli:
> I personally do not like `black` too much but I do not have much against it either. I believe auto formatting code does not need manual review. This might mean that a separate PR could be used to house all changes ensuing from running a formatter against the codebase.

I agree. @cmeesters I think a major code refactor, even if it only changes the code formatting, must go in a separate PR. Besides that, I think it can be worth deciding the code style we'd like to use for this project so we don't need to refactor in the future (and of course linting and/or blacking every PRs).

cmeesters:
@ashleycaselli there are several issues. I overstepped the boundary of a single fix because they are interacting:
- a missing version to be able to use poetry for testing locally, which also can be used within the CI

For pyproject.toml we have:
- the rdflib dependency is outdated
- "coveralls" was not listed
- the project title change is not considered
- I deliberately changed to the poetry format, as it is the more common one

Then, the CI makes use of `isort`, which throws numerous errors. I (successfully) tried to embed linting and formatting into the CI workflow, instead of invoking shell scripts. I used flake8 and black.

I included a coverage run - which I think was missing.

There were some version conflicts and a Pythonversion, which was too old. 

The rest is minor. By including black, I made too many changes. I did not anticipate, that black could be disliked. I really sometimes hate it myself. Yet, it really helps to improve the readability. In the end, I changed too many files. Particularly by running black.

In the end, only the linting failed, which could be configured away, as this was nitpicking on the tests, only. Otherwise, this PR displayed a working CI.

All in all, this is an impressive and useful library. I will depend on it for future projects. Hence, the need of a working auto-release.

ashleycaselli:
@cmeesters thanks for the detailed reply!

Indeed, some points have been fixed, while others are still WIP or have not even started yet. I will open issues for each group of changes, so there is more granularity and control over the changes. Feel free to comment or open new issues if those that I opened are not enough/complete.

PR #233, opened 2026-06-17
Support `~~~ARTIFACTCODE~~~` placeholder in custom namespaces (#232)

### LINKED ISSUE #232: Support `~~~ARTIFACTCODE~~~` placeholder for custom-namespace artifact codes
## Summary

Support the `~~~ARTIFACTCODE~~~` placeholder so the trusty artifact code can be minted into a **custom namespace**, not only as a sub-URI of the nanopub itself.

Example goal — sign a nanopub that introduces a concept in a project namespace:

- nanopub URI stays standard: `https://w3id.org/np/RA<code>`
- introduced thing gets the **same** code in another namespace: `https://w3id.org/peh/biochementities/RA<code>`

This is already implemented in the Java library (`org.nanopub.trusty` / `net.trustyuri`, which recognises `~~~ARTIFACTCODE~~~`, with the deprecated `ARTIFACTCODE-PLACEHOLDER` fallback, matched as `~~~ARTIFACTCODE~~~[\.#/]?$`), but is missing here.

## Current behaviour

The vendored trustyuri only rewrites URIs under the dummy namespace `http://purl.org/nanopub/temp/np/`. In `nanopub/trustyuri/rdf/RdfUtils.py::get_trustyuri`, any URI outside that namespace hits `suffix is None → return str(resource)` and is passed through untouched — so a placeholder in a custom namespace survives verbatim into the signed output.

Repro (v2.1.0):

```python
import rdflib
from nanopub import Profile, NanopubConf, Nanopub
from rdflib.namespace import RDFS

term = rdflib.URIRef("https://w3id.org/peh/biochementities/~~~ARTIFACTCODE~~~")
a = rdflib.Graph()
a.add((term, RDFS.label, rdflib.Literal("test")))
prof = Profile(orcid_id="https://orcid.org/0000-0002-1825-0097", name="Bot")
conf = NanopubConf(profile=prof, use_test_server=True,
                   add_prov_generated_time=True, attribute_publication_to_profile=True)
np = Nanopub(conf=conf, assertion=a)
np.sign()
print("~~~ARTIFACTCODE~~~" in np.rdf.serialize(format="trig"))  # -> True (placeholder NOT substituted)
```

## Proposed fix

`get_trustyuri` is the single choke point used by both the hashing/verification pass and the final replacement pass (via `RdfPreprocessor.transform`), so the change is localized there:

1. Substitute `~~~ARTIFACTCODE~~~` → `hashstr` in **any** URI (handles sign-hashing where `hashstr=" "`, and final replacement where `hashstr` is the artifact code).
2. On **verification** — when `base_uri` itself carries this nanopub's `RA…` code — blank that same code wherever it appears in out-of-namespace URIs. Without this, the resolved thing URI is hashed as literal content and `has_valid_trusty` fails.
3. Leave references to *other* trusty URIs (different `RA…` codes) untouched.

A ~10-line proof-of-concept wrapping `get_trustyuri` along these lines produces: the introduced thing URI carrying the **same** code as the nanopub, the placeholder gone, references to other biochementities preserved, and both `has_valid_trusty` and `has_valid_signature` true.

## Cross-implementation parity

This behaviour is not currently covered by `nanopub-testsuite` (no `transform/` case uses a custom-namespace artifact code). I've opened Nanopublication/nanopub-testsuite#3 to add such a case so Java/Python/JS/Rust can be checked for parity (same input → same artifact code).

Happy to open a PR with the fix + a unit test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

tkuhn:
Fixes #232.

## Problem

The nanopub testsuite added a `transform/artifactcode-1` case that signs a concept declared in a **custom namespace** using the `~~~ARTIFACTCODE~~~` placeholder:

```turtle
this:assertion {
  <https://example.org/ns/~~~ARTIFACTCODE~~~> a ex:Concept ; ...
}
```

The expected signed output gives that concept the nanopub's own trusty code (`ex:RAK4W62...`). Python's vendored trustyuri left any URI outside the nanopub's namespace untouched, so the placeholder survived signing, the hash didn't match, and `test_testsuite_sign_valid[rsa-key1/artifactcode-1.in.trig]` failed.

## Fix

Following the nanopub-java / trustyuri-java behaviour, `get_trustyuri` now handles URIs outside the nanopub's own namespace in three scenarios:

1. **Signing/hashing** — substitute `~~~ARTIFACTCODE~~~` with the hash placeholder so the concept is part of the hash.
2. **Applying the trusty artifact** — the same substitution writes the real trusty code into the custom namespace.
3. **Verification** — blank *this* nanopub's own trusty code back to the hash placeholder so the hash matches, while leaving references to *other* trusty nanopubs intact.

The produced trusty code is byte-identical to the Java reference output (`RAK4W62QQBgpEOgpDfW3bldxScwlIuWOmAkOL8xhSfxeQ`).

## Changes

- `nanopub/definitions.py` — add `ARTIFACTCODE_PLACEHOLDER` constant.
- `nanopub/trustyuri/rdf/RdfUtils.py` — substitute/blank the placeholder and code in the out-of-namespace branch.
- `tests/test_sign_utils.py` — add a network-independent regression test.

## Testing

All 425 tests pass (`uv run pytest`), including the previously failing testsuite case and the new regression test.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/tests/test_sign_utils.py
+++ b/tests/test_sign_utils.py
@@ -7,7 +7,68 @@ from nanopub.definitions import NP_PREFIX, NP_TEMP_PREFIX
 from nanopub.namespaces import NPX
 from nanopub.sign_utils import add_signature
 from nanopub.trustyuri.rdf import RdfUtils
-from tests.conftest import profile_test
+from tests.conftest import profile_test, testsuite_conf
+
+
+ARTIFACTCODE_PLAIN_TRIG = """\
+@prefix this: <http://purl.org/nanopub/temp/1029384756/> .
+@prefix np: <http://www.nanopub.org/nschema#> .
+@prefix dct: <http://purl.org/dc/terms/> .
+@prefix npx: <http://purl.org/nanopub/x/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+@prefix rdfs: <http://www.w3.org/2000/01/rdf-schema#> .
+@prefix orcid: <https://orcid.org/> .
+@prefix prov: <http://www.w3.org/ns/prov#> .
+@prefix ex: <https://example.org/ns/> .
+
+this:Head {
+  this: a np:Nanopublication;
+    np:hasAssertion this:assertion;
+    np:hasProvenance this:provenance;
+    np:hasPublicationInfo this:pubinfo .
+}
+
+this:assertion {
+  <https://example.org/ns/~~~ARTIFACTCODE~~~> a ex:Concept;
+    rdfs:label "a concept minted in a custom namespace" .
+}
+
+this:provenance {
+  this:assertion prov:wasAttributedTo orcid:0000-0002-1267-0234 .
+}
+
+this:pubinfo {
+  this: dct:created "2024-09-18T11:19:06.183Z"^^xsd:dateTime;
+    dct:creator orcid:0000-0002-1267-0234;
+    npx:introduces <https://example.org/ns/~~~ARTIFACTCODE~~~> .
+}
+"""
+
+
+def test_sign_artifactcode_placeholder_in_custom_namespace():
+    """Signing a nanopub that mints a concept in a custom namespace via the
+    ``~~~ARTIFACTCODE~~~`` placeholder gives that concept the same trusty code
+    as the nanopub itself, and the result is valid (see issue #232)."""
+    ds = Dataset()
+    ds.parse(data=ARTIFACTCODE_PLAIN_TRIG, format="trig")
+    np = Nanopub(conf=testsuite_conf, rdf=ds)
+    np.sign()
+
+    assert np.has_valid_signature
+    assert np.has_valid_trusty
+    assert np.is_valid
+
+    artifact_code = str(np.source_uri).removeprefix(NP_PREFIX)
+    assert re.fullmatch(r"RA[A-Za-z0-9\-_]{43}", artifact_code)
+
+    # The placeholder must be gone and the concept must carry the nanopub's code.
+    concept = URIRef(f"https://example.org/ns/{artifact_code}")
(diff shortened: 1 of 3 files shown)


VirginiaBalseiro (approved):

tests/test_sign_utils.py
ashleycaselli, line 1: For consistency, I'd move these tests in this class: https://github.com/Nanopublication/nanopub-py/blob/af7afef1a87696c16ecc64a3fdfc089c98c4a565/tests/test_nanopub.py#L394

Also, I'd read the nanopub from the `nanopub-testsuite` project instead of defining it inline.

tkuhn: There is a new test case on this in the testsuite. Is it fine to have this test here too, or better remove it here entirely?

ashleycaselli: the `test_sign_utils.py` only contains the test cases of the [`sign_utils.py`](https://github.com/Nanopublication/nanopub-py/blob/main/nanopub/sign_utils.py). The test cases concerning the signing process are contained under the [`TestSign`](https://github.com/Nanopublication/nanopub-py/blob/af7afef1a87696c16ecc64a3fdfc089c98c4a565/tests/test_nanopub.py#L394) class.

In addition, to reduce redundancy and possible errors, instead of defining the TRIG as a plain String, I would suggest to use the testsuite-connector to get the TRIG content of the specific nanopub file (e.g., through `testsuite.get_by_nanopub_uri(...)`).

tkuhn: Thanks for the feedback! Addressed in follow-up #234: moved the test to the `TestSign` class in `test_nanopub.py` and now read the input nanopub from the testsuite via the connector instead of an inline TRIG string.

PR #241, opened 2026-06-29
fix(profile): accept standard PEM/DER RSA keys on load

VirginiaBalseiro:
Profile / np setup only accepted nanopub's own bare-base64 key format. A standard PEM private key (what openssl/ssh-keygen produce) failed with an opaque `binascii.Error: Incorrect padding`, with no hint at the cause.

This PR fixed that by adding:

 * Normalize any RSA key (PKCS# 1/# 8 PEM, DER, CRLF line endings, or bare base64) through RSA.import_key -> single-line base64
 * Raise a clear ProfileError on unparseable input. 
 * Regression tests for PKCS#1, PKCS#8, and CRLF PEM.
 
Also: 

* Consolidated key encoding behind shared internal helpers (_encode_private_key / _encode_public_key), used by both the generate and load paths so there's a single definition of nanopub's canonical key format. format_key() is now deprecated and will be removed in an upcoming major version.

### DIFF
--- a/tests/test_nanopub.py
+++ b/tests/test_nanopub.py
@@ -392,6 +392,65 @@ class TestCreationFromTrustyNanopub:
         assert np.get_source_uri_from_graph == "http://purl.org/np/RA1sViVmXf-W2aZW4Qk74KTaiD9gpLBPe2LhMsinHKKz8"
 
 
+class TestProfileKeyFormats:
+    """Loading a Profile from a key file should accept a standard PEM key.
+
+    Regression test for the opaque ``binascii.Error: Incorrect padding`` raised
+    when ``np setup`` (or ``Profile(..., private_key=Path)``) is pointed at a
+    normal PEM private key. The read path at profile.py does ``f.read().strip()``
+    and then ``decodebytes(...)`` without stripping the ``-----BEGIN-----`` armor
+    or newlines, so anything that isn't nanopub's own bare single-line base64
+    blows up deep inside base64 instead of loading or failing with a clear error.
+    """
+
+    @staticmethod
+    def _write_pem(tmp_path, pkcs):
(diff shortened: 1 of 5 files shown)


tkuhn:
What's important is that we don't allow for any variation in the pubkey literal in the nanopub's pubinfo. Because that is used to do exact matching in SPARQL queries.

VirginiaBalseiro:
> What's important is that we don't allow for any variation in the pubkey literal in the nanopub's pubinfo. Because that is used to do exact matching in SPARQL queries.

Covered. The pubinfo value is exactly Literal(profile.public_key), and for keys already in canonical form normalization is a byte-for-byte identity (verified, plus unchanged trusty-URI hashes). Added a regression test to lock it.

tests/test_nanopub.py
```diff
@@ -392,6 +392,65 @@ def test_get_source_uri_from_graph_returns_trusty(self, testsuite):
...
+        return key_file, pem
+
+    def test_standard_pkcs1_pem_key_loads(self, tmp_path):
+        """A PKCS#1 PEM key (-----BEGIN RSA PRIVATE KEY-----) should load."""
+        from nanopub.profile import Profile
+
+        key_file, pem = self._write_pem(tmp_path, pkcs=1)
+        assert pem.startswith("-----BEGIN RSA PRIVATE KEY-----")
+
+        profile = Profile("0000-0000-0000-0000", "Tester", private_key=key_file)
```
ashleycaselli, line 424: Currently, this shouldn't be allowed. `orcid_id` must always be a URI. Either we raise an exception, or we implement it to accept also the `id` at creation time.
 
> profile = Profile("0000-0000-0000-0000", "Tester", private_key=key_file)

ashleycaselli (changes requested):

ashleycaselli (approved):

PR #243, opened 2026-07-07
Allow signer id to be any URI (agent_id)

### LINKED ISSUE #242: Allow signer ID to be any valid URI

VirginiaBalseiro:
Closes #242 .

This PR generalizes the ORCID-only signer to a agent_id that accepts any URI at face value (an ORCID, another identity provider, or a bot/agent URI). A bare ORCID (0000-0000-0000-0000) still expands to its canonical URI, and only values claiming to be an ORCID (https://orcid.org/… prefix) are format-checked. We don't do any extra validation (such as verifying that the URI is dereferenceable, etc) in this PR, only basic regex validation.

Backwards-compatible (non-breaking) changes:

* orcid_id kept as a deprecated alias (constructor kwarg + property; warns on the Python API only).
* Legacy orcid_id: key in profile.yml still loads; store() now writes agent_id:.
* CLI setup/sign accept any URI, format-checking only ORCID-shaped values.
* Removal of the deprecated orcid_id is left for a future major.

This PR also adds tests for accepting any URI, bare-ORCID expansion, malformed-ORCID rejection, deprecated alias, and legacy yaml load.

### DIFF
--- a/examples/create_and_publish.ipynb
+++ b/examples/create_and_publish.ipynb
@@ -12,13 +12,7 @@
    "cell_type": "markdown",
    "id": "b9b5555b-b43c-4d9f-8d36-33e34c4868d8",
    "metadata": {},
-   "source": [
-    "For the code below to work, you need to have your profile set up, specifying your ORCID identifier and generating a key pair, using this command:\n",
-    "\n",
-    "    $ np setup\n",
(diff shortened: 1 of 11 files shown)


tkuhn (approved):
looks good!

PR #250, opened 2026-08-06
fix: reject ill-typed literals during validation

### LINKED ISSUE #249: Ill-typed literals are published without warning
## Summary

Literals whose value does not match their declared datatype (e.g. `"not-a-number"^^xsd:integer`) are signed and published without any warning. Strict RDF stores reject such nanopubs, so they end up published but invisible in the SPARQL endpoint — the failure mode described in [nanopub-java#12](https://github.com/Nanopublication/nanopub-java/issues/12).

Checked against `main` at 4500a89.

## Details

- rdflib does not catch this. `Literal('not-a-number', datatype=XSD.integer)` constructs without error (`.value` is `None`) and parses fine from N-Triples — verified locally.
- `Nanopub.is_valid` only checks that the four graphs are non-empty, distinct, and correctly named. It never inspects literals.
- `Nanopub.publish()` does not call `is_valid` at all — it goes straight to `sign()` and then `publish_graph()`.

The original trigger of nanopub-java#12 is already fixed here: `nanopub.py` writes `Literal(datetime.now().astimezone(), datatype=XSD.dateTime)`, so the integer-typed `dct:created` is no longer generated. What remains is that an ill-typed literal in a *user-supplied* triple passes through silently.

## Suggested fix

`pyproject.toml` requires `rdflib >= 6.3.2`, which has [`Literal.ill_typed`](https://rdflib.readthedocs.io/en/stable/apidocs/rdflib.html#rdflib.term.Literal.ill_typed). A loop over the dataset in `is_valid` flagging `o.ill_typed` covers this in a few lines, with no new dependency:

```python
for s, p, o, c in self._rdf.quads():
    if isinstance(o, Literal) and o.ill_typed:
        raise MalformedNanopubError(f"Literal {o!r} is not valid for its datatype")
```

Worth deciding separately whether `publish()` should call `is_valid` first, since today no validation runs on the publishing path.

## Context

The equivalent check was added to nanopub-java in [Nanopublication/nanopub-java#93](https://github.com/Nanopublication/nanopub-java/pull/93), using RDF4J's `XMLDatatypeUtil.isValidValue`, reported through `NanopubVerifier` and enforced by `PublishNanopub --strict`. nanopub-js has the same gap: [Nanopublication/nanopub-js#15](https://github.com/Nanopublication/nanopub-js/issues/15).

ashleycaselli:
Closes #249

## Problem

Literals whose lexical form does not match their declared datatype (e.g. `"not-a-number"^^xsd:integer`) were signed and published without any warning. rdflib constructs them happily (`.value` is `None`) and parses them back from TriG, and `Nanopub.is_valid` only ever inspected the graph structure. Strict RDF stores reject such nanopubs, so they end up published but invisible in the SPARQL endpoint — the failure mode of [nanopub-java#12](https://github.com/Nanopublication/nanopub-java/issues/12).

## Changes

- **`Nanopub._check_ill_typed_literals()`** — walks every quad and raises `MalformedNanopubError` for object literals with `ill_typed` true, reporting all offenders in one message (each in `n3()` form with its graph) rather than failing on the first. Uses `rdflib.term.Literal.ill_typed`, available since rdflib 6.3.2, which is already the floor in `pyproject.toml` — no new dependency.
- **`is_valid`** calls it, after the structural checks and before signature/trusty verification, so `sign()` and the `nanopub check` CLI command both catch it.
- **`publish()`** now validates nanopubs that were already signed. Previously validation ran only as a side effect of `sign()`, so a nanopub loaded from disk or fetched and then republished skipped it entirely.

Literals with no datatype, with a language tag, or with an unrecognized datatype IRI have `ill_typed = None` and pass untouched — only recognized XSD datatypes with a bad lexical form are rejected.

## Tests

New `TestIllTypedLiterals` covers rejection in the assertion graph, in pubinfo, and through the TriG parse path (confirming the flag survives a serialization round-trip); acceptance of well-typed and untypeable literals; `sign()` refusing to sign; and `publish()` refusing an already-signed nanopub without calling `publish_graph`. Full suite passes (503 tests).

## One thing worth a second opinion

`is_valid` also runs at construction time for any trusty nanopub (`nanopub.py:141`), so **fetching** an already-published nanopub that contains an ill-typed literal now raises instead of loading — and that would also block retracting one. Such nanopubs exist in the wild; that is the premise of nanopub-java#12. This PR follows the fix suggested in the issue. If we would rather mirror nanopub-java (report on read, enforce on publish), the change is to log a warning at that construction call site and keep the raise for the sign/publish path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/autorelease.yml
+++ b/.github/workflows/autorelease.yml
@@ -21,6 +21,7 @@ jobs:
         rdflib-version:
           - rdflib==6.3.2
           - rdflib==7.1.4
+          - rdflib==7.6.0
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -16,6 +16,7 @@ jobs:
         rdflib-version:
           - rdflib==6.3.2
           - rdflib==7.1.4
+          - rdflib==7.6.0
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         with:
--- a/nanopub/__main__.py
+++ b/nanopub/__main__.py
@@ -129,6 +129,9 @@ def check(filepath: Path):
     try:
         np.is_valid
         print(f"\033[1m✅ Valid nanopub\033[0m {np.source_uri}")
+        for literal, graph in np.ill_typed_literals:
+            print(f"\033[1m⚠️  Ill-typed literal\033[0m {literal.n3()} in graph {graph}: "
+                  "it cannot be signed or published as is")
     except MalformedNanopubError as e:
         print(f"\033[1m❌ Invalid nanopub\033[0m: {e}")
 
--- a/nanopub/__main__.py
+++ b/nanopub/__main__.py
@@ -131,7 +131,7 @@ def check(filepath: Path):
         print(f"\033[1m✅ Valid nanopub\033[0m {np.source_uri}")
         for literal, graph in np.ill_typed_literals:
             print(f"\033[1m⚠️  Ill-typed literal\033[0m {literal.n3()} in graph {graph}: "
-                  "it cannot be signed or published as is")
+                  "strict RDF stores may be missing this nanopub")
     except MalformedNanopubError as e:
         print(f"\033[1m❌ Invalid nanopub\033[0m: {e}")
 
--- a/nanopub/nanopub.py
+++ b/nanopub/nanopub.py
@@ -244,7 +244,10 @@ class Nanopub:
     def publish(self) -> Tuple[str, str, str | None]:
         """Publish a Nanopub object"""
         if not self.source_uri:
+            # sign() validates the nanopub before signing it
             self.sign()
+        elif not self.is_valid:
+            raise MalformedNanopubError("The nanopub is not valid, cannot publish it")
 
         publish_graph(self.rdf, use_server=self._conf.use_server)
         logger.info(f'Published {self.source_uri} to {self._conf.use_server}')
@@ -342,6 +345,8 @@ class Nanopub:
         if not found_pubinfo:
             raise MalformedNanopubError(
                 f"The pubinfo graph should contain at least one triple that has the nanopub URI as subject: \033[1m{self._source_uri}\033[0m")
+
+        self._check_ill_typed_literals()
 
         if self._metadata.signature:
             if not self.has_valid_signature:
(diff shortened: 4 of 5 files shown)


ashleycaselli:
The CI failure (`tests/test_testsuite.py::test_testsuite_valid_trusty[fair-maturity-1.trig]`) turned out to be the exact tradeoff flagged in the description, so I have moved enforcement to the write path in 97b3f48.

**What the failure showed.** `fair-maturity-1.trig` is a real published nanopub that the test suite lists as *valid*, and its pubinfo carries `"2019-02-26"^^xsd:dateTime` — a date-only lexical form for a datatype that requires the time part. Rejecting it in `is_valid` diverged from the reference test suite and would have made such nanopubs unreadable (and unretractable).

It passed locally and failed in CI purely because of rdflib: 6.3.2 parses `xsd:dateTime` with `isodate`, which is strict, while 7.6 delegates to `datetime.fromisoformat`, which accepts a date-only string on Python 3.12. So `ill_typed` is `True` on 6.3.2 and `False` on 7.6 for the very same literal — another reason not to hinge readability on it.

**What changed.**

- `is_valid` logs a warning instead of raising, so existing nanopubs stay readable and verifiable.
- `sign()` and `publish()` refuse — the actual bug in #249 is a user-supplied ill-typed triple getting signed and published.
- New public `Nanopub.ill_typed_literals` property returning `(literal, graph)` pairs, so callers can check without triggering an exception.
- `np check` prints a `⚠️ Ill-typed literal` line per offender alongside the `✅ Valid nanopub` verdict — the same shape as nanopub-java's `NanopubVerifier` reporting while `--strict` enforces.

This matches the nanopub-java split: report on read, enforce on publish.

**Verified** against all three rdflib versions — 6.3.2 (the job that failed), 7.1.4, and 7.6.0 — 504 tests passing on each.

tkuhn:
This makes a lot of sense to me. We should probably also have this in the test suite and implement it for the other nanopub libraries.

I suppose this can only check a small number of general datatypes, not user-defined ones, obviously?

ashleycaselli:
> This makes a lot of sense to me. We should probably also have this in the test suite and implement it for the other nanopub libraries.

It's already implemented in `nanopub-java` in https://github.com/Nanopublication/nanopub-java/pull/93. `nanopub-js` has an open issue for that (https://github.com/Nanopublication/nanopub-js/issues/15).

> I suppose this can only check a small number of general datatypes, not user-defined ones, obviously?

Yes, currently it only works with general datatypes because it's using the `rdflib` checker. For the custom-defined ones, we must create our own checker.

